Issue #10, opened 2024-05-29
Add a security policy


ahmedsobeh:
The repo is missing a security policy that is consistent with that of valkey

PR #11, opened 2024-05-29
adding security policy

ahmedsobeh:
### Pull Request check-list

- [ ] Was the change added to CHANGES file?

### Description of change

This PR adds a security policy to the repo. One that is consistent with that of Valkey itself

This PR resolves issue #10

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,7 @@
+## Reporting a Vulnerability
+
+If you believe you've discovered a security vulnerability, please contact the Valkey team at security@lists.valkey.io.
+Please *DO NOT* create an issue.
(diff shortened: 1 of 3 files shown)


SECURITY.md
```diff
@@ -0,0 +1,7 @@
+## Reporting a Vulnerability
```
aiven-sal, line 1: LGTM
but we should also update this:

https://github.com/valkey-io/valkey-py/blob/main/CONTRIBUTING.md#security-vulnerabilities

maybe we should just refer to SECURITY.md

ahmedsobeh: ah true, I updated it to be consistent with contributing.md

aiven-sal (approved):
Thanks!

Issue #23, opened 2024-06-26
Provide support for redis:// and rediss://


kjaymiller:
https://github.com/valkey-io/valkey-py/blob/d116aa6efa03877a0c37598139a556554ad47c13/valkey/asyncio/connection.py#L1025-L1074

This states that only `valkey://`, `valkeys://`, and `unix` but some providers like aiven are still using the `rediss://`.

Is there a reason that we aren't allowing the `redis://` and `rediss://` protocols or at least providing a deprecation warning if they are used?

This error gave me initial pause on if I could use the `valkey` client to test with new providers of Valkey.

kjaymiller:
also happy to contribute a PR that adds redis:// and rediss://.

aiven-sal:
Thanks for reporting this issue!

You can use `valkey://` and `valkeys://` in place of `redis://` and `rediss://` regardless of what your provider says you should be using.
Valkey and Redis protocol are the same at the moment and, if they will ever change, valkey-py will only support the Valkey protocol.

But I agree that the transition will be easier if we supported redis:// and rediss:// as aliases of valkey:// and valkeys://, as long as the two protocols remain identical.

Of course PRs are always welcome.

kjaymiller:
happy to contribute. I think the biggest question to add would be should a warning be issued when `redis://` or `rediss://` are used?

aiven-sal:
I think that, at this point, it would be nice if people could swap redis-py with valkey-py without noticing any difference at all. So I wouldn't add a warning. At some point in the future we should definitely add a deprecation warning, but for now I'd like to make the transition as seamless as possible.

aiven-sal:
Fixed by #29

Issue #25, opened 2024-06-27
implement a single validation function around the service_uri to reduce code/test complexity around connection URL Parsing


kjaymiller:
Thanks for wanting to report an issue you've found in valkey-py. Please delete this text and fill in the template below.  
It is of course not always possible to reduce your code to a small test case, but it's highly appreciated to have as much data as possible. Thank you!

**Version**:  5.1.0b7

**Platform**: Python 3.12 on MacOS ARM running OS Sonoma

**Description**: 

In #23 I proposed including `redis` and `rediss` as valid protocols. I noticed while implementing the change that the code was almost identical with the main difference being variable names.

![Diffs of code base showing that the code in both sync and async versions are the same](https://github.com/valkey-io/valkey-py/assets/8632637/e8ee98c0-6672-426e-a272-e07031f22ca5)

I propose that the protocol and url code be moved into a module that can be called in both places. This would reduce identical code reducing the amount of code needed to cover as well as making the testing for that portion of the code-base completely non-reliant on server setup and connection. This tests that the code does what it's supposed to do not that the code interacts with the platform how we expect it to (which should be a different and separate test in itself)

Proposed changes:
- move parse_url to a new python module (or possibly refactor the aysnc connection to inherit from the synchronous one if possible)
- create a dedicated test path for that parsing and move the identical tests in the sync/async versions of `test_connection_pool.py.TestConnectionPoolURLParsing` to that dedicated test module.

aiven-sal:
It definitely makes sense to share `parse_url` implementation across the 2 classes. And we should keep the "asyncio" version, because the other is actually broken (it checks the URI scheme case sensitively which is against RFC3986). 
It would certainly make sense to test the "new" parse_url in a single place, but I would still keep the separate TestConnectionPoolURLParsing (at least part of them) because they don't just test `parse_url`.

PR #27, opened 2024-06-27
adds protocol support for `redis` and `rediss`

### LINKED ISSUE #26: missing pytest.mark.skipif on TestSSLConnectionURLParsing
**Version**: 5.1.0b7

**Platform**: Python 3.12 / MacOS Sonoma in M1 Arm Device

**Description**: 

There is an inconsistent skip on testing if SSL isn't installed between the connection pool in the sync and async tests.

These tests are otherwise identical (pointing to my suggested issue #25)

sync:
https://github.com/valkey-io/valkey-py/blob/d116aa6efa03877a0c37598139a556554ad47c13/tests/test_connection_pool.py#L456C1-L494C1

async:
https://github.com/valkey-io/valkey-py/blob/d116aa6efa03877a0c37598139a556554ad47c13/tests/test_asyncio/test_connection_pool.py#L543-L572

Unless there is a reason the async client should not skip the tests (or a particular reason the synchronous one should) is it not better to make the tests test the same thing the same way (or implement #25 and condense that logic into a module that both versions use which would eliminate the need for testing in both areas.

kjaymiller:
### Pull Request check-list

- [X] Was the change added to CHANGES file?

### Description of change

- Adds 'redis://' and 'rediss://' to the list of accepted protocols for both sync and async connection pools
- Updates tests to use parameterization so that tests for valkey:// and valkeys:// also test redis// and rediss://

resolves #26 #23

### DIFF
--- a/tests/test_connection_pool.py
+++ b/tests/test_connection_pool.py
@@ -196,7 +196,6 @@ class TestBlockingConnectionPool:
         expected = "path=abc,db=0,client_name=test-client"
         assert expected in repr(pool)
 
-
 @pytest.mark.parametrize("connection_protocol", ["valkey", "redis"])
 class TestConnectionPoolURLParsing:
     def test_hostname(self, connection_protocol):
--- a/tests/test_connection_pool.py
+++ b/tests/test_connection_pool.py
@@ -196,6 +196,7 @@ class TestBlockingConnectionPool:
         expected = "path=abc,db=0,client_name=test-client"
         assert expected in repr(pool)
 
+
 @pytest.mark.parametrize("connection_protocol", ["valkey", "redis"])
 class TestConnectionPoolURLParsing:
     def test_hostname(self, connection_protocol):
@@ -204,7 +205,9 @@ class TestConnectionPoolURLParsing:
         assert pool.connection_kwargs == {"host": "my.host"}
 
     def test_quoted_hostname(self, connection_protocol):
-        pool = valkey.ConnectionPool.from_url(f"{connection_protocol}://my %2F host %2B%3D+")
+        pool = valkey.ConnectionPool.from_url(
+            f"{connection_protocol}://my %2F host %2B%3D+"
+        )
         assert pool.connection_class == valkey.Connection
         assert pool.connection_kwargs == {"host": "my / host +=+"}
 
@@ -215,7 +218,9 @@ class TestConnectionPoolURLParsing:
 
     @skip_if_server_version_lt("6.0.0")
     def test_username(self, connection_protocol):
-        pool = valkey.ConnectionPool.from_url(f"{connection_protocol}://myuser:@localhost")
+        pool = valkey.ConnectionPool.from_url(
+            f"{connection_protocol}://myuser:@localhost"
+        )
         assert pool.connection_class == valkey.Connection
         assert pool.connection_kwargs == {"host": "localhost", "username": "myuser"}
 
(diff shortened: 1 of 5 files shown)


valkey/asyncio/connection.py
```diff
@@ -1023,6 +1023,20 @@ class ConnectKwargs(TypedDict, total=False):
 
 
 def parse_url(url: str) -> ConnectKwargs:
+    valid_schemes = (
```
aiven-sal, line 1026: This check shouldn't be here and, in fact, it should be removed from the other `parse_url` as well. Schemes in URIs should not be case sensitive (cfr. [RFC3986 Section 3.1](https://datatracker.ietf.org/doc/html/rfc3986#section-3.1))

aiven-sal (changes requested):
Thanks for the PR!
I have a couple of requests:
1) could you squash the lint changes in your second commit in the first one? It would make it easier for me to review commit-by-commit.
2) ignore the CHANGES file, it wasn't really used much recently and we have a nice bot that can summarize changes in releases from PRs. It isn't perfect, but for now let's just use that.

valkey/asyncio/connection.py
kjaymiller: @aiven-sal - regarding raising that error, should a similar check be done to the `ParseResult.scheme` which would return the protocol in all lowercase.

```py
# an example
>>> url =urlparse("HTTPX://foo.com:900")
>>> url
ParseResult(scheme='httpx', netloc='foo.com:900', path='', params='', query='', fragment='')
```

aiven-sal: I think you can just leave the check as it was before (adding redis/rediss of course) with the exception raised in the `else`, so that we know for sure that what we check is what we use and we don't end up with 2 different places to maintain/update with the supported schemas.

tests/test_connection_pool.py
```diff
@@ -286,67 +301,72 @@ def test_extra_typed_querystring_options(self):
...
+
+def test_invalid_scheme_raises_error_when_double_slash_missing():
+    with pytest.raises(ValueError) as cm:
+        valkey.ConnectionPool.from_url("valkey:foo.bar.com:12345")
+    assert str(cm.value) == (
+        "Valkey URL must specify one of the following schemes "
+        "(valkey://, valkeys://, redis://, rediss://, unix://)"
+    )
+
+
```
kjaymiller, line 350: These tests (and their async component) would need to be removed as there is no check at this point.

```suggestion
```

valkey/asyncio/connection.py
kjaymiller: @aiven-sal - I agree however the sync version wasn't checking in that manner. Should I apply the same check to the sync version?

aiven-sal: That can be done in a different PR (or maybe not needed at all if we unify the two parse_url).
Let's jut focus on adding redis/rediss.

kjaymiller:
> 1. could you squash the lint changes in your second commit in the first one? It would make it easier for me to review commit-by-commit.

I hope I did this correctly

valkey/asyncio/connection.py
kjaymiller: I've added back the original test at the end and removed the check in sync (and it's test since there is no check on sync still)

kjaymiller:
@aiven-sal - Is there something on my end holding progress on this?

aiven-sal:
> @aiven-sal - Is there something on my end holding progress on this?

Nope, there is some related work going on in https://github.com/valkey-io/valkey-py/pull/29, so I'm waiting for it to get merged. Then this will need to be rebased (I'm sorry for the inconvenience).

aiven-sal:
So #29 was merged. It should handle #23 and #25 . 
This PR also contains the fix for #26.
To keep things tidier I'll close this PR, so you can open another one just for the #26 fix.
Thanks again and sorry for the delay. It was just easier to fix #23 and #25 together.

PR #29, opened 2024-06-30
 Implement a single validation function around the service_uri

ahmedsobeh:
### Pull Request check-list

- [ ] Was the change added to CHANGES file?

### Description of change

Created url_parser and removed duplication of parse_url. This PR resolves issue https://github.com/valkey-io/valkey-py/issues/25

I created a url_parser module and added it to the parsers folder. I then used this module in both connection modules (the normal one and the asyncio one) and updated the tests accordingly.

### DIFF
--- a/valkey/_parsers/url_parser.py
+++ b/valkey/_parsers/url_parser.py
@@ -1,0 +1,104 @@
+from types import MappingProxyType
+from typing import Callable, Mapping, Optional
+from urllib.parse import ParseResult, parse_qs, unquote, urlparse
+
+from valkey.asyncio.connection import ConnectKwargs
+from valkey.asyncio.connection import SSLConnection as SSLConnectionAsync
+from valkey.asyncio.connection import (
+    UnixDomainSocketConnection as UnixDomainSocketConnectionAsync,
+)
+from valkey.connection import SSLConnection, UnixDomainSocketConnection
+
+
+def to_bool(value) -> Optional[bool]:
+    if value is None or value == "":
+        return None
+    if isinstance(value, str) and value.upper() in FALSE_STRINGS:
+        return False
+    return bool(value)
+
+
+FALSE_STRINGS = ("0", "F", "FALSE", "N", "NO")
+
+URL_QUERY_ARGUMENT_PARSERS: Mapping[str, Callable[..., object]] = MappingProxyType(
+    {
+        "db": int,
+        "socket_timeout": float,
+        "socket_connect_timeout": float,
+        "socket_keepalive": to_bool,
+        "retry_on_timeout": to_bool,
+        "max_connections": int,
+        "health_check_interval": int,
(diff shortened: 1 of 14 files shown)


valkey/_parsers/url_parser.py
```diff
@@ -0,0 +1,104 @@
...
+                    raise ValueError(f"Invalid value for `{name}` in connection URL.")
+            else:
+                kwargs[name] = value
+
+    if parsed.username:
+        kwargs["username"] = unquote(parsed.username)
+    if parsed.password:
+        kwargs["password"] = unquote(parsed.password)
+
+    # We only support valkey://, valkeys:// and unix:// schemes.
```
mkmkme, line 70: Should we also support redis and rediss, as requested in other issue?

ahmedsobeh: It depends on what gets merged first 😅  If this PR gets merged first, then updating the branch of #27 with main's latest would cover it. if #27 gets merged first, I'll update my PR accordingly and it should cover it too

ahmedsobeh: on second thought, I'll just add it now to reduce conflicts

aiven-sal:
LGTM, but CI is a bit cranky. Will merge later.

valkey/_parsers/url_parser.py
```diff
@@ -0,0 +1,100 @@
...
+        "timeout": float,
+    }
+)
+
+
+def parse_url(url: str, async_connection: bool):
+    url_prefixes = ["valkey://", "valkeys://", "unix://", "redis://", "rediss://"]
+    supported_schemes = ["valkey", "valkeys", "redis", "rediss"]
+
+    if not any(url.startswith(prefix) for prefix in url_prefixes):
```
aiven-sal, line 42: The exception is already raised on line 96. This `if` is redundant and it is also case-sensitive and it shouldn't be.
We can just drop the whole `if` block.

ahmedsobeh: ah that's true. removed it.

ahmedsobeh: turns out this actually covers the test **_test_invalid_scheme_raises_error_when_double_slash_missing_**. Without this, not having a double slash doesn't raise an error.

aiven-sal: Oh right. I think the best way to check for that would be something like:
```
if not parsed.netloc:
    raise something
```

ahmedsobeh: done ✅

ahmedsobeh: doesn't really work with unix, so I added a regex for it

aiven-sal (approved):
LGTM
I squashed and rebased some commits and will merge as soon as the CI completes

PR #40, opened 2024-07-10
Make pytest pass locally with Python 3.12

### LINKED ISSUE #24: Add development support for python 3.12
Thanks for wanting to report an issue you've found in valkey-py. Please delete this text and fill in the template below.  
It is of course not always possible to reduce your code to a small test case, but it's highly appreciated to have as much data as possible. Thank you!

**Version**: What valkey-py and what valkey version is the issue happening on?

**Platform**: Python 3.12 on MacOS (M-Series ARM devices)

**Description**: There are some package changes that need to be made to make development possible on python 3.12

- unpinning the version of invoke from 1.7.3 allowing version >=2.0.0 which removes usage of the `imp` module which was removed in Python 3.12
- replacing `distutils` which is being used to check for truthy-like strings using `strtobool` with a dedicated module still being maintained like `dirty-equals`

mkmkme:
### Pull Request check-list

### Description of change

This is the first PR addressed to fix #24. It contains a smaller amount of changes, but each change is complicated enough to move this bunch into a separate PR.

There will be a second one, fixing calls for `invoke` and some style issues.

These commits fix missing dependencies and deadlocks.

### DIFF
--- a/tests/test_asyncio/test_connect.py
+++ b/tests/test_asyncio/test_connect.py
@@ -166,12 +166,13 @@ async def _assert_connect(
         await aserver.start_serving()
         try:
             await conn.connect()
+            stop_event.set()
             await conn.disconnect()
         except ConnectionError:
             finished.set()
             raise
         finally:
-            stop_event.set()
+            stop_event.set()  # Set stop_event in case of a connection error
             aserver.close()
             await aserver.wait_closed()
             await finished.wait()
--- a/tests/test_connect.py
+++ b/tests/test_connect.py
@@ -95,7 +95,13 @@ def test_tcp_ssl_tls12_custom_ciphers(tcp_address, ssl_ciphers):
         ssl_min_version=ssl.TLSVersion.TLSv1_2,
         ssl_ciphers=ssl_ciphers,
     )
-    _assert_connect(conn, tcp_address, certfile=certfile, keyfile=keyfile)
+    _assert_connect(
+        conn,
+        tcp_address,
+        certfile=certfile,
+        keyfile=keyfile,
+        ssl_version=ssl.TLSVersion.TLSv1_2,
+    )
 
 
 @pytest.mark.ssl
@@ -118,7 +124,7 @@ def test_tcp_ssl_version_mismatch(tcp_address):
             tcp_address,
             certfile=certfile,
             keyfile=keyfile,
-            ssl_version=ssl.PROTOCOL_TLSv1_2,
+            ssl_version=ssl.TLSVersion.TLSv1_3,
         )
 
 
@@ -147,7 +153,7 @@ class _ValkeyTCPServer(socketserver.TCPServer):
         *args,
         certfile=None,
         keyfile=None,
-        ssl_version=ssl.PROTOCOL_TLS,
+        ssl_version=ssl.TLSVersion.TLSv1,
         **kw,
     ) -> None:
         self._ready_event = threading.Event()
@@ -174,12 +180,12 @@ class _ValkeyTCPServer(socketserver.TCPServer):
         if self._certfile is None:
             return super().get_request()
         newsocket, fromaddr = self.socket.accept()
-        connstream = ssl.wrap_socket(
+        sslctx = ssl.create_default_context(ssl.Purpose.CLIENT_AUTH)
+        sslctx.load_cert_chain(self._certfile, self._keyfile)
+        sslctx.minimum_version = self._ssl_version
+        connstream = sslctx.wrap_socket(
             newsocket,
             server_side=True,
-            certfile=self._certfile,
-            keyfile=self._keyfile,
-            ssl_version=self._ssl_version,
         )
         return connstream, fromaddr
 
--- a/valkey/commands/graph/query_result.py
+++ b/valkey/commands/graph/query_result.py
@@ -1,6 +1,5 @@
 import sys
 from collections import OrderedDict
-from distutils.util import strtobool
 
 # from prettytable import PrettyTable
 from valkey import ResponseError
@@ -37,6 +36,22 @@ STATS = [
     CACHED_EXECUTION,
     INTERNAL_EXECUTION_TIME,
 ]
+
+
+def strtobool(value: str) -> bool:
+    """
+    Convert a string representation of truth to True or False.
+    True values are 'y', 'yes', 't', 'true', 'on', and '1'; false values
+    are 'n', 'no', 'f', 'false', 'off', and '0'.  Raises ValueError if
+    'value' is anything else.
+    """
+    value = value.lower()
+    if value in ("y", "yes", "t", "true", "on", "1"):
+        return True
+    elif value in ("n", "no", "f", "false", "off", "0"):
+        return False
+    else:
+        raise ValueError(f"invalid truth value {value}")
 
 
 class ResultSetColumnTypes:
@@ -270,7 +285,7 @@ class QueryResult:
         """
         value = value.decode() if isinstance(value, bytes) else value
         try:
-            scalar = True if strtobool(value) else False
+            scalar = strtobool(value)
         except ValueError:
             sys.stderr.write("unknown boolean type\n")
             scalar = None
--- a/tests/test_asyncio/test_cluster.py
+++ b/tests/test_asyncio/test_cluster.py
@@ -60,6 +60,7 @@ class NodeProxy:
         self.send_event = asyncio.Event()
         self.server = None
         self.task = None
+        self.pipes = None
         self.n_connections = 0
 
     async def start(self):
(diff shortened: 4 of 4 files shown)


tests/test_connect.py
```diff
@@ -174,12 +174,12 @@
         if self._certfile is None:
             return super().get_request()
         newsocket, fromaddr = self.socket.accept()
-        connstream = ssl.wrap_socket(
+        sslctx = ssl.create_default_context(ssl.Purpose.CLIENT_AUTH)
+        sslctx.load_cert_chain(self._certfile, self._keyfile)
+        sslctx.check_hostname = False
+        connstream = sslctx.wrap_socket(
```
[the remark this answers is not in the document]
mkmkme: Gonna fix this by adding missing `min_version` and `max_version`.

Not critical though, as appears only in tests

valkey/commands/graph/query_result.py
```diff
@@ -39,6 +38,22 @@
 ]
 
 
+def strtobool(value: str) -> bool:
```
aiven-sal, line 41: This new version of the function allows you to change

https://github.com/valkey-io/valkey-py/pull/40/files#diff-24a05b46c3b945ca8d3369e3f0195f190695945f6f77ed6a8c0c3c996cf40aa9L273

to 

```scalar = strtobool(value)```

File: tests/test_connect.py (hunk 1 above)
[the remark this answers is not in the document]
Comment (aiven-sal) [reply tests/test_connect.py:180]: It's okay. You can leave it this way. It's just a test.

File: tests/test_asyncio/test_connect.py [hunk 3]
```diff
@@ -189,7 +189,7 @@ async def _valkey_request_handler(reader, writer, stop_event):
         except TimeoutError:
             continue
         if not buffer:
-            continue
+            break
```
aiven-sal, line 192: Can you explain why this change is needed?
Looks like the behavior of valkey-py changed here. Is that so?

mkmkme: It's only the behaviour of test, but yeah it has changed here. I described it in the commit message:

>Helper `_valkey_request_handler` had a logical mistake in case `Reader`
didn't contain any data. In that case, the handler would endlessly loop
and `await` on `Reader.read()` preventing the `ioloop` from iterating
further. This used to work in Python 3.9, but it seems that the
implementation of `ioloop` is stricter in Python 3.12 which caused the
inner loop of `_valkey_request_handler` to never end.
>
>The proper way of handling the case when there's no data in `Reader` is
to quit the handler, close `Writer` and let `ioloop` progress.
Therefore, simply replacing `continue` with `break` fixes the issue.

aiven-sal: What I mean is that I'd like to better understand why this loop loops forever in 3.12.
Because I would expect other people doing this kind of stuff to use "continue" instead of "break".
In the real world (over a real network connection) it isn't that unusual to get and empty read from an open socket. 
And one shouldn't simply kill a connection the first time a read returns nothing.
The thing that is making this test fail could cause problems to users using valkey in a similar way.

mkmkme: Sure, I'll happily explain myself here. When it comes to `async` part, things start getting really hairy :D

I might be wrong with an assessment that "ioloop implementation is stricter in Python 3.12" but it's certainly different, from what I see when comparing logs from 3.9 and 3.12. 3.9 actually looks like it performs everything synchronously.

I need to put a stress on one thing here: `writer.close()` here does not kill the whole connection. Instead, it finishes the handler iteration, allowing the event loop to proceed. Next time the data appears in `Reader`, handler will be called again and will have to drain the whole buffer. This cycle will continue until `Reader` gets closed.
Closing `Writer` here means we don't have anything to send _for now_.

For instance, take a look at the TCP echo server using streams in Python documentation: https://docs.python.org/3/library/asyncio-stream.html#tcp-echo-server-using-streams

Here, it also closes the writer as soon as it drains the reader. The connection to the other endpoint persists (due to `serve_forever`.

I hope I'm making any sense with this :D Please tell me if something is left unanswered.

mkmkme: Also, just to back up my words, I verified that we're getting exactly the same data with both `break` and `continue` on Python 3.9. You can trigger this by invoking `pytest --log-cli-level=INFO  --log-format="%(filename)s:%(lineno)d: %(asctime)s %(levelname)s %(message)s"  --log-date-format=" %Y-%m-%d %H:%M"  --protocol=2 -W always -m 'not onlycluster' tests/test_asyncio/test_connect.py  -x`

```
test_connect.py:186:  2024-07-10 14:40 INFO False
test_connect.py:197:  2024-07-10 14:40 INFO Command fragment: *3
test_connect.py:197:  2024-07-10 14:40 INFO Command fragment: $6
test_connect.py:197:  2024-07-10 14:40 INFO Command fragment: CLIENT
test_connect.py:197:  2024-07-10 14:40 INFO Command fragment: $7
test_connect.py:197:  2024-07-10 14:40 INFO Command fragment: SETNAME
test_connect.py:197:  2024-07-10 14:40 INFO Command fragment: $17
test_connect.py:197:  2024-07-10 14:40 INFO Command fragment: test-suite-client
test_connect.py:217:  2024-07-10 14:40 INFO Command CLIENT SETNAME test-suite-client
test_connect.py:219:  2024-07-10 14:40 INFO Response from b'+OK\r\n'
test_connect.py:186:  2024-07-10 14:40 INFO False
test_connect.py:197:  2024-07-10 14:40 INFO Command fragment: *4
test_connect.py:197:  2024-07-10 14:40 INFO Command fragment: $6
test_connect.py:197:  2024-07-10 14:40 INFO Command fragment: CLIENT
test_connect.py:197:  2024-07-10 14:40 INFO Command fragment: $7
test_connect.py:197:  2024-07-10 14:40 INFO Command fragment: SETINFO
test_connect.py:197:  2024-07-10 14:40 INFO Command fragment: $8
test_connect.py:197:  2024-07-10 14:40 INFO Command fragment: LIB-NAME
test_connect.py:197:  2024-07-10 14:40 INFO Command fragment: $9
test_connect.py:197:  2024-07-10 14:40 INFO Command fragment: valkey-py
test_connect.py:217:  2024-07-10 14:40 INFO Command CLIENT SETINFO LIB-NAME valkey-py
test_connect.py:219:  2024-07-10 14:40 INFO Response from b'-ERR\r\n'
test_connect.py:197:  2024-07-10 14:40 INFO Command fragment: *4
test_connect.py:197:  2024-07-10 14:40 INFO Command fragment: $6
test_connect.py:197:  2024-07-10 14:40 INFO Command fragment: CLIENT
test_connect.py:197:  2024-07-10 14:40 INFO Command fragment: $7
test_connect.py:197:  2024-07-10 14:40 INFO Command fragment: SETINFO
test_connect.py:197:  2024-07-10 14:40 INFO Command fragment: $7
test_connect.py:197:  2024-07-10 14:40 INFO Command fragment: LIB-VER
test_connect.py:197:  2024-07-10 14:40 INFO Command fragment: $7
test_connect.py:197:  2024-07-10 14:40 INFO Command fragment: 5.1.0b7
test_connect.py:217:  2024-07-10 14:40 INFO Command CLIENT SETINFO LIB-VER 5.1.0b7
test_connect.py:219:  2024-07-10 14:40 INFO Response from b'-ERR\r\n'
test_connect.py:186:  2024-07-10 14:40 INFO False
test_connect.py:223:  2024-07-10 14:40 INFO Exit handler
PASSED
```

On Python 3.12 with `continue`, the first line (`test_connect.py:186:  2024-07-10 14:40 INFO False`) was looping forever due to test awaiting on the same _state_ of `reader`  because it was stuck in handler

aiven-sal: > I need to put a stress on one thing here: `writer.close()` here does not kill the whole connection. Instead, it finishes the handler iteration, allowing the event loop to proceed. Next time the data appears in `Reader`, handler will be called again and will have to drain the whole buffer. This cycle will continue until `Reader` gets closed. Closing `Writer` here means we don't have anything to send _for now_.
> 
> For instance, take a look at the TCP echo server using streams in Python documentation: https://docs.python.org/3/library/asyncio-stream.html#tcp-echo-server-using-streams

The link you posted says that `write.close()` closes the socket (This is more explicit https://docs.python.org/3/library/asyncio-stream.html#asyncio.StreamWriter.close ). 
I think we may be talking about different things here.
I'm thinking about a single command split over multiple reads with an empty read in the middle for some reason. Not about different connections/clients.
So when I say "close the socket" I mean close this specific socket connected to this specific client, not kill the entire server and the listening socket.

But let's just put this discussion aside for a moment. 
What I'm arguing is that I would expect the loop to loop forever _also_ in 3.9, if it wasn't for that `stop_event`.
So if 3.12 is looping forever, it's probably because `stop_event` is not working, not because of the continue. If the continue was a problem it would be a problem in any version, wouldn't it?

mkmkme: (Just doubling here what already said in the DM)

Yup, this does look like something I've overlooked. Thanks for the catch! I'll dig further :)

mkmkme: Fixed in https://github.com/valkey-io/valkey-py/pull/40/commits/6866ee79d915f69805117ca8625109c86a078a77

aiven-sal: Thanks for looking into this hairy issue!
I leave this link here for future reference: https://github.com/python/cpython/issues/104344

aiven-sal (approved):

PR #32, opened 2024-07-03
Replace hiredis dependency with libvalkey

mkmkme:
**NOTE:** This PR should not be merged until `libvalkey-py` is moved into `valkey-io` and the first version of `libvalkey-py` is released to PyPI!

### Pull Request check-list

_Please make sure to review and check all of these items:_

- [X] Do tests and lints pass with this change?
- [ ] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)?
- [X] Is the new or changed code fully tested?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [ ] Is there an example added to the examples folder (if applicable)?
- [ ] Was the change added to CHANGES file?

_NOTE: these things are not required to open a PR and can be done
afterwards / while the PR is open._

### Description of change

This changeset changes hiredis dependency with libvalkey.

PR #52, opened 2024-07-12
Fix RESP3 CI

### LINKED ISSUE #44: RESP3 tests limited scope
RESP3 tests don't run against some python versions and some tests are excluded.

We should run RESP3 tests against the same python version used by the "regular" workflow and we should not exclude any test.

ahmedsobeh:
### Pull Request check-list

### Description of change

This PR changes the RESP3 CI tests to run on the same python versions as the other tests. It also removed exclusions that were previously there. It would resolve #44


aiven-sal:
Interesting... apparently the reason why RESP3 tests didn't run with hiredis option is that it segfaults. I wonder if this happen with libvalkey-py as well. I might be worth waiting for #32 and see if anything improves.

Issue #46, opened 2024-07-11
Unify RESP3 and other tests workflows


aiven-sal:
Right now GH action workflow for RESP3 and everything else are separated.
We should create a single workflow for both using the `matrix` feature.
So we will be sure that we won't miss anything in the tests (as it happens now).

Before we can do this we need to fix:

 - [x] #44 
 - [x] #45

PR #69, opened 2024-08-05
Unify RESP3 and other tests workflows

ahmedsobeh:
### Pull Request check-list

### Description of change
This PR removes RESP3 tests flow and combines it with run-tests. This would resolbe #46

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -60,9 +60,10 @@ jobs:
          python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', 'pypy-3.9', 'pypy-3.10']
          test-type: ['standalone', 'cluster']
          connection-type: ['libvalkey', 'plain']
+         protocol-version: ['2','3']
      env:
        ACTIONS_ALLOW_UNSECURE_COMMANDS: true
-     name: Python ${{ matrix.python-version }} ${{matrix.test-type}}-${{matrix.connection-type}} tests
+     name: Python ${{ matrix.python-version }} RESP${{ matrix.protocol-version }} ${{matrix.test-type}}-${{matrix.connection-type}} tests
      steps:
        - uses: actions/checkout@v4
        - uses: actions/setup-python@v5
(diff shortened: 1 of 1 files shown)


.github/workflows/integration.yaml
```diff
@@ -60,9 +60,10 @@ jobs:
          python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', 'pypy-3.9', 'pypy-3.10']
          test-type: ['standalone', 'cluster']
          connection-type: ['libvalkey', 'plain']
+         protocol: ['--protocol=3', '']
```
mkmkme, line 63: Maybe add `--protocol=3` explicitly here?

ahmedsobeh: I did it in a different way, take another look

.github/workflows/integration.yaml
```diff
@@ -79,9 +80,16 @@ jobs:
            fi
            invoke devenv
            sleep 10 # time to settle
-           invoke ${{matrix.test-type}}-tests
-           if [[ "${{matrix.python-version}}" != pypy-* ]]; then
-            invoke ${{matrix.test-type}}-tests --uvloop
+           if [[ "${{matrix.protocol}}" == RESP3 ]]; then
+            invoke ${{matrix.test-type}}-tests --protocol=3
+            if [[ "${{matrix.python-version}}" != pypy-* ]]; then
+            invoke ${{matrix.test-type}}-tests --uvloop --protocol=3
```
mkmkme, line 86: Two nitpicks and one question:

**Nitpick 1:** Indentation of `if` end `else` branches are different, please use two spaces everywhere
**Nitpick 2:** Missing indentation in inner `if`

**Question:** Wouldn't it be easier to have something like this

```
protocol-version: ['2', '3']
name: ... RESP${{matrix.protocol-version}}
...
invoke ${{matrix.test-type}}-tests --protocol=${{matrix.protocol-version}}
```
? You'd get rid of one additional `if` statement

ahmedsobeh: That looks better, I was just trying to keep the original naming of tests.

mkmkme: Ah I see.

I don't know. For me, the explicit protocol version in the test name looks more handy. It might not be obvious for everyone which RESP version is used by default.

ahmedsobeh: yep, I did it that way now 👍

ahmedsobeh: I ran a yaml formatter on the file, so you might see some additional changes.

mkmkme: Tests seem to be failing. I guess because `PyPy` is also called just `Python` now?

mkmkme: Also I'd personally prefer the yaml formatter to be on a separate commit/PR but I guess it's fine

mkmkme (dismissed):
LGTM! Let's see what tests are saying.

Also it seems that semi-linear history check is failing already now. You might have to rebase it on the top of `main`

PR #70, opened 2024-08-06
Import some changes from redis-py

aiven-sal:
### Description of change

* Fix socket_timeout init
* Add missing type hints for `retry.py`
* Add details to the asyncio connection error message
* Format connection errors in the same way everywhere
* Drop unused dev-dep urllib3
* Close Unix sockets if the connection attempt fails
* Cluster performance improvements
* Bump version deps
* Fix checking of module versions
* Avoid dangling sockets in case of SSL handshake failure
* Get rid of event_loop warnings
* Resolve some docs warnings
* Ensure safe defaults for TLS

### DIFF
--- a/tests/test_asyncio/test_connection.py
+++ b/tests/test_asyncio/test_connection.py
@@ -494,3 +494,21 @@ async def test_connection_garbage_collection(request):
 
     await client.aclose()
     await pool.aclose()
+
+
+@pytest.mark.parametrize(
+    "error, expected_message",
+    [
+        (OSError(), "Error connecting to localhost:6379. Connection reset by peer"),
+        (OSError(12), "Error connecting to localhost:6379. 12."),
+        (
+            OSError(12, "Some Error"),
+            "Error 12 connecting to localhost:6379. [Errno 12] Some Error.",
+        ),
+    ],
+)
+async def test_connect_error_message(error, expected_message):
+    """Test that the _error_message function formats errors correctly"""
+    conn = Connection()
+    error_message = conn._error_message(error)
+    assert error_message == expected_message
--- a/tests/test_asyncio/test_connection.py
+++ b/tests/test_asyncio/test_connection.py
@@ -14,7 +14,11 @@ from valkey._parsers import (
     parse_url,
 )
 from valkey.asyncio import ConnectionPool, Valkey
-from valkey.asyncio.connection import Connection, UnixDomainSocketConnection
+from valkey.asyncio.connection import (
+    Connection,
+    SSLConnection,
+    UnixDomainSocketConnection,
+)
 from valkey.asyncio.retry import Retry
 from valkey.backoff import NoBackoff
 from valkey.exceptions import ConnectionError, InvalidResponse, TimeoutError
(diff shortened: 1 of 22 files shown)


tests/test_asyncio/test_connection.py
mkmkme, line 1: Here  "Cu-authored-by:" also shows you and not the original author

tests/test_connection.py
```diff
@@ -291,3 +291,53 @@ def mock_disconnect(_):
 
     assert called == 1
     pool.disconnect()
+
+
+@pytest.mark.parametrize(
+    "conn, error, expected_message",
+    [
```
mkmkme, line 298: I wonder if it makes sense to move this collection out to some shared variable to additionally check it's the same in async and sync cases. But it's minor nonetheless, feel free to ignore if you want.

valkey/connection.py
mkmkme, line 1: Co-authored-by here as well :)

tests/test_retry.py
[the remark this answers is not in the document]
aiven-sal: I'm not sure which commit to are referring to

tests/test_asyncio/test_connection.py
aiven-sal: As above, I'm not sure which commit to are referring to

valkey/connection.py
aiven-sal: As Above, I'm not sure which commit to are referring to

mkmkme:
Yep sorry didn't know it wouldn't show you. I'm talking about commits:
* 6236f51119ecc4a55a1b5318f6fe7871d63f41d8
* 337a7ee641a18873bfa7f00e59686a4c667b96fa
* 1889058211c3aa8b4f1e7016f121844a44586c34
* af92e15d69da815f2b0625bbab4be3b591ef19ca

aiven-sal:
> Yep sorry didn't know it wouldn't show you. I'm talking about commits:
> 
>     * [6236f51](https://github.com/valkey-io/valkey-py/commit/6236f51119ecc4a55a1b5318f6fe7871d63f41d8)
> 
>     * [337a7ee](https://github.com/valkey-io/valkey-py/commit/337a7ee641a18873bfa7f00e59686a4c667b96fa)
> 
>     * [1889058](https://github.com/valkey-io/valkey-py/commit/1889058211c3aa8b4f1e7016f121844a44586c34)
> 
>     * [af92e15](https://github.com/valkey-io/valkey-py/commit/af92e15d69da815f2b0625bbab4be3b591ef19ca)

I added the Co-authored-by tag because those commits aren't copied verbatim. There are some minor changes in the code or the commit message compared to the original.
Of course I kept the original commit author, but I did want to unambiguously acknowledge that the code isn't exactly the one they wrote.
Maybe I'm overthinking it, but I have seen issues arising in other forks of other projects because of similar situations, with people threatening to sue etc

tests/test_connection.py
aiven-sal: It would make sense, but it's more work that what it looks.
The collection contains `SSLConnection` but it's not the same `SSLConnection`. There are 2 different `SSLConnection` in async and non-async code.
I guess one day we will de-duplicate a lot of that code. But for now I think it's better to no go down that rabbit hole.

mkmkme:
> > Yep sorry didn't know it wouldn't show you. I'm talking about commits:
> > ```
> > * [6236f51](https://github.com/valkey-io/valkey-py/commit/6236f51119ecc4a55a1b5318f6fe7871d63f41d8)
> > 
> > * [337a7ee](https://github.com/valkey-io/valkey-py/commit/337a7ee641a18873bfa7f00e59686a4c667b96fa)
> > 
> > * [1889058](https://github.com/valkey-io/valkey-py/commit/1889058211c3aa8b4f1e7016f121844a44586c34)
> > 
> > * [af92e15](https://github.com/valkey-io/valkey-py/commit/af92e15d69da815f2b0625bbab4be3b591ef19ca)
> > ```
> 
> I added the Co-authored-by tag because those commits aren't copied verbatim. There are some minor changes in the code or the commit message compared to the original. Of course I kept the original commit author, but I did want to unambiguously acknowledge that the code isn't exactly the one they wrote. Maybe I'm overthinking it, but I have seen issues arising in other forks of other projects because of similar situations, with people threatening to sue etc

Thanks, now I understand :) Just was wondering whether this was intentional

mkmkme (approved):

PR #77, opened 2024-08-23
Fix issue with pytest-asyncio v0.24

aiven-sal:
### Description of change

Drop decorator mark.asyncio as it wasn't doing anything here since pytest-asyncio v0.6 and now its unused
parameter has been dropped.


mkmkme (approved):
LGTM, thank you!

PR #86, opened 2024-09-04
ci: cache docker layers and avoid re-downloading them for each job

aiven-sal:
### Description of change
 
Sometimes the CI fails because Docker doesn't respond.
It could be because we download many docker images in every job and we have 58 jobs now.
With this commit, the image layers will be downloaded during nightly builds and cached.


mkmkme (approved):
LGTM!

PR #88, opened 2024-09-09
tests: increase fudge factors

aiven-sal:
### Description of change

Make some tests less likely to fail.
Maybe we should just drop these tests.
But let's see if we can make them less flaky.

### DIFF
--- a/tests/test_asyncio/test_lock.py
+++ b/tests/test_asyncio/test_lock.py
@@ -107,7 +107,7 @@ class TestLock:
     async def test_blocking_timeout(self, r):
         lock1 = self.get_lock(r, "foo")
         assert await lock1.acquire(blocking=False)
-        bt = 0.2
+        bt = 0.9
         sleep = 0.05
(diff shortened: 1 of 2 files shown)


mkmkme (approved):
I wonder if those factors are important. Like, if there are situations when we do have to do something in a certain amount of time and no longer.

If you're positive these changes won't hide any bug in valkey-py, I'm fine with this.

aiven-sal:
You can't guarantee that a blocking operation with a timeout will not take more than timeout+delta, unless you have an RTOS and you are not using python. If you rely on the fact that this timeouts will return "on time" you already have a bug in your code.

In fact, I'm not sure it makes any sense at all to check how much time passed in those tests, but if this commit fixes the flaky tests we can leave them.
If they fail again, I'll just drop them completely.
The actual amount of time that those operation take is completely outside of our control and influenced by stuff (i.e. other processes running on the machine) that have nothing to do with valkey-py.

PR #87, opened 2024-09-08
update python version constraint to < 3.11 for async-timeout requirement 

ali-gholami-aiven:
async-timeout library has effectively been upstreamed into Python 3.11+. installing it for python >= 3.11 should not be necessary.

this will enable packaging valkey-py for fedora 39, which comes with python 3.12 and does not conatin async-timeout 4.0.3 package.

see the deprecation notice:
https://github.com/aio-libs/async-timeout?tab=readme-ov-file#deprecated

### Pull Request check-list

- [ ] Do tests and lints pass with this change?
- [ ] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)?
- [ ] Is the new or changed code fully tested?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [ ] Is there an example added to the examples folder (if applicable)?

PR #92, opened 2024-09-10
Make valkey-py install the correct dependencies in Python 3.11.0 to Python 3.11.3

nirav-ark-biotech:
I just happened to be using Python 3.11.1 while using valkey-py. 

The PR #87 broke my CI/CD pipeline (when upgrading to valkey-py `6.1` bc the codebase, in https://github.com/valkey-io/valkey-py/blob/03320e6b2720a90d1b2ac7946d82bdc4d87f3faa/valkey/asyncio/connection.py#L31 

just so happens to check whether or not we're 3.11.3 and above.

If you are in Python 3.11.0 - 3.11.2, you go into the else and this code fails if async timeout is not installed

so we need to still to install `async-timeout` if python version < 3.11.3

### Pull Request check-list

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,7 +36,7 @@ setup(
     author_email="valkey-py@lists.valkey.io",
     python_requires=">=3.8",
     install_requires=[
-        'async-timeout>=4.0.3; python_version<"3.11"',
+        'async-timeout>=4.0.3; python_version<"3.11.3"',
     ],
     classifiers=[
         "Development Status :: 5 - Production/Stable",
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -78,7 +78,7 @@ jobs:
        max-parallel: 15
        fail-fast: false
        matrix:
-         python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', 'pypy-3.9', 'pypy-3.10']
+         python-version: ['3.8', '3.9', '3.10', '3.11', '3.11.1', '3.12', 'pypy-3.9', 'pypy-3.10']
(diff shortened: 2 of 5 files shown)


mkmkme (approved):
Thanks for your contribution! It looks good to me.

Two questions:
1. Did you verify locally that this does fix your problem?
2. Could you squash the commits into one as the second one is a fixup to the first one?

aiven-sal:
Hi Nirav! Thanks for reporting the problem!
The issue is totally valid and should be fixed ASAP.
But I think that a better fix would be to replace the 
```
if sys.version_info >= (3, 11, 3):
```
with
```
if sys.version_info >= (3, 11):
```
Can you amend your PR?

Apologies for the inconvenience, btw

aiven-sal:
Hi!
Thanks for reporting and apologies for the inconvenience.
I think that setup.py would also need to be updated.
I'll add a commit here for that, so we can make a new release quickly

setup.py
```diff
@@ -36,7 +36,7 @@
     author_email="valkey-py@lists.valkey.io",
     python_requires=">=3.8",
     install_requires=[
-        'async-timeout>=4.0.3; python_version<"3.11"',
+        'async-timeout>=4.0.3; python_version<"3.11.4"',
```
mkmkme, line 39: Should be `<3.11.3` here

mkmkme (changes requested):

mkmkme (dismissed):
LGTM thanks!

ahmedsobeh (commented):
LGTM 🚀

aiven-sal (approved):

mkmkme (approved):

Issue #84, opened 2024-09-02
[Feature request] Typing support and migrating from redis-py


Rafiot:
**Version**: 6.0

**Platform**: Any

**Description**: The typing provided by the redis package [is terrible](https://github.com/redis/redis-py/issues/2399#issuecomment-2298091739) and the recommended way is to use [types-redis](https://pypi.org/project/types-redis/) but this one obviously doesn't work with the valkey package, and there is no `types-valkey` (yet?).

I'm playing around to see if I can find a janky solution, but is there any chance you're planning to merge (and support) the types hints directly in the package? Or provide a `types-valkey` package at some point?

aiven-sal:
Hi!
I'd say that ideally we would like to have proper typing support. Of course adding types in a project of this size isn't a simple task. Re-using the work done for "types-redis" could certainly help.
Have you use that? Does it work well with recent versions of redis-py? It seems to be stuck at 4.6

Rafiot:
Hi! Yes, I'm using it with redis-py 5.0.8 and it works perfectly fine, but I might very well not be using new features that are unsupported (https://github.com/python/typeshed/issues/10592)

Rafiot:
Alright, so this is going to be a fun project. I may tell you things you already know, but that was new to me and there are a few questions to answer before starting to work on that.

First of all, as far as I can tell, the typing hints are completely broken and never work for anyone:

#### Async

```python
import asyncio

from valkey.asyncio import Valkey


async def test() -> dict[bytes, bytes]:
    v: Valkey = Valkey()
    response_hset: int = await v.hset('foo', mapping={'bar': 'baz'})
    print(response_hset)
    h: dict[bytes, bytes] = await v.hgetall('foo')
    print(h)
    return h

asyncio.run(test())
```

```bash
$ (git::main) mypy test_async.py 
test_async.py:10: error: Incompatible types in "await" (actual type "Awaitable[int] | int", expected type "Awaitable[Any]")  [misc]
test_async.py:12: error: Incompatible types in "await" (actual type "Awaitable[dict[Any, Any]] | dict[Any, Any]", expected type "Awaitable[Any]")  [misc]
Found 2 errors in 1 file (checked 1 source file)
```

#### Sync

``` python
from valkey import Valkey

v: Valkey = Valkey()
v.hset('foo', mapping={'bar': 'baz'})
h: dict[bytes, bytes] = v.hgetall('foo')
print(h)

```

```
$ (git::main) mypy test.py 
[...]

PR #93, opened 2024-09-11
Types

Rafiot:
### Pull Request check-list


Rafiot:
Woops, I wanted to do a PR on my own repo, but mixed it up and did the PR here, sorry for that.

It is not ready for merging, but I guess as it is here now, feel free to comment.

(WIP for https://github.com/valkey-io/valkey-py/issues/84)

mkmkme:
Sure. As I mentioned in the issue, I'd really like to see the types within the code and not in separate files.

Also please you `--signoff` when committing changes (we can fix that later with the commits that are already here)

Rafiot:
I'll also squash the commits, it is a mess for no good reason right now.

Rafiot:
Now a note that is possibly a bit of a problem: `smembers` is marked as returning a set, but it doesn't. It always returns a list.

That's the reason the tests are failing there : https://github.com/Rafiot/valkey-py/actions/runs/10808666835/job/29982264205?pr=1#step:6:300

But also, the test suite is very much not helping as it sometimes casts the list to a set, when len > 1 (again, see test suite above).

**Question**: Should we cast the response of smembers to a set? Or change the type hints?

aiven-sal:
> Question: Should we cast the response of smembers to a set? Or change the type hints?

We should change the type hints. There is some difference between RESP3 sets and Python's sets and it is not possible to represents all RESP3 sets as sets in Python, so the functions has to return a list. Its up to the user to cast it to set when/if they know that is is safe to do so.

Rafiot:
smsmbers is now marked as returning a list

mkmkme:
>smsmbers is now marked as returning a list

@Rafiot I noticed the last commit changes some test code as well. That's a bit odd. Is your branch based on some old commit in `valkey-py`? Or is there other reason to change the test?

In my understanding, this PR should not change any behaviour of the library, it only adds the type hints.

Rafiot:
@mkmkme I'm using the test suite as a way to validate the type hints, so to do that, I need to add type hints to the test suite, which explains the changes. I'll review it, but it should not change the logic of the tests just add the hints in a way mypy is happy. It seemed to me to be the best way to make sure the type hints are in-line with the current codebase. 
Would you have another approach?

mkmkme:
@Rafiot no-no, this approach sounds correct. I was just surprised by the commit [2e504dd](https://github.com/valkey-io/valkey-py/pull/93/commits/2e504dde8cf877d69a276be0f711fab9ac0f5f6d)

Rafiot:
Oh, true. So I'm confused how the tests as they are now can pass, because the response is definitely a list, not a set:

```
>>> v = Valkey()
>>> v.ping()
True
>>> v.sadd('foo', 'bar')
1
>>> v.smembers('foo')
[b'bar']
>>> v.smembers('blah')
[]
>>> v.sadd('foo', 'baz')
1
>>> v.smembers('foo')
[b'bar', b'baz']
```

Rafiot:
I just want to quickly update you on that one: I haven't forgotten, and will get back to it asap. This month got busy but I should be able to get back to this PR in a few days.

Rafiot:
I did terrible things to this branch. New PR: https://github.com/valkey-io/valkey-py/pull/134

PR #100, opened 2024-09-24
tasks: added --pytest-args to tests

### LINKED ISSUE #62: allow for `TESTARGS` in invoke test
**Version**: 6.0.0b1

**Platform**:  MacOS Arm M2

**Description**: Description of your issue, stack traces from errors and code that reproduces the issue

Currently the `invoke test` command pre-configures some commands but pushes you down a few paths:

1. Test all the tests in standalone
2. Test all the tests in a cluster
3. Test all the tests in both

I propose that we add the following to all the tests functions in `task.py` either:

1. add a `TESTARGS` parameters that allows for the user to apply their own args to the test
2. add individual parameters for things like `test_path`, `-v (verbosity)`, `-k (keyword)`

mkmkme:
### Pull Request check-list

### Description of change

Adds support for arbitrary pytest args for `invoke tests/standalone_tests/cluster_tests`

Fixes #62

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -28,7 +28,9 @@ def build_docs(c):
 def linters(c, color=False):
     """Run code linters"""
     run(f"flake8 --color {'always' if color else 'never'} tests valkey")
-    run(f"black {'--color' if color else ''} --target-version py37 --check --diff tests valkey")
+    run(
+        f"black {'--color' if color else ''} --target-version py37 --check --diff tests valkey"
+    )
     run(f"isort {'--color' if color else ''} --check-only --diff tests valkey")
     run("vulture valkey whitelist.py --min-confidence 80")
     run("flynt --fail-on-change --dry-run tests valkey")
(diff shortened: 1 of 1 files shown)


tasks.py
```diff
@@ -42,40 +44,43 @@ def all_tests(c, color=False):
...
 
 
 @task
-def cluster_tests(c, uvloop=False, protocol=2, color=False):
+def cluster_tests(c, uvloop=False, protocol=2, color=False, pytest_args=None):
     """Run tests against a valkey cluster"""
     cluster_url = "valkey://localhost:16379/0"
+    color_arg = f"--color={'yes' if color else 'no'}"
+    shared_args = f"{color_arg} --protocol={protocol} --cov=./ --cov-report=xml:coverage_cluster.xml -W always -m 'not onlynoncluster and not valkeymod' --valkey-url={cluster_url}"
+    pytest_args = pytest_args or ""
```
aiven-sal, line 77: Should we somehow try to validate `pytest_args` ?
As it is, it can be used to inject any kind of command (I don't mean just any arguments, you can use it to run any executable).
We, probably, shouldn't be _too_ concerned about a malicious actor slipping some bad command in this argument, as it seems an unlikely scenario. But it may be possible for someone to inadvertently do something bad/wrong. Especially since complex argument passing will require _a lot_ of escaping to be done correctly in this way.
I'm not sure if it's possible to sanitize this command in any reasonable way returning intelligible errors (most likely not). Maybe it would be better to handle the main pytest arguments as separate args here.

tasks.py
```diff
@@ -28,7 +28,9 @@ def build_docs(c):
 def linters(c, color=False):
     """Run code linters"""
     run(f"flake8 --color {'always' if color else 'never'} tests valkey")
-    run(f"black {'--color' if color else ''} --target-version py37 --check --diff tests valkey")
+    run(
```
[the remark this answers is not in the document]
mkmkme: Yeah autoformatting changed this line for me :/ Do you want me to move it to a separate commit?

tasks.py
```diff
@@ -42,40 +44,43 @@ def all_tests(c, color=False):
 
 
 @task
-def tests(c, uvloop=False, protocol=2, color=False):
+def tests(c, uvloop=False, protocol=2, color=False, pytest_args=None):
     """Run the valkey-py test suite against the current python,
     with and without libvalkey.
     """
     print("Starting Valkey tests")
+    print("pytest_args:", pytest_args)
```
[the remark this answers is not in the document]
mkmkme: This is actually a leftover that I forgot to remove. Would you rather see it removed or having an additional proper message about pytest args?

tasks.py
mkmkme: We can do this, yes. This approach leaves it to user's discretion. After all, any arguments they would pass they would run on *their* machine :)

aiven-sal: What I mean is that this argument seem to be not easy to use, but it's easy to misuse.
Not easy to use: because you need to put your desired list of arguments in quotes but the most typical argument you may want to pass (i.e. -k) often needs quotes itself, so you need to use the right type of quotes or use escapes.
Easy to misuse: because I wouldn't expect a command like `invoke tests --pytest-args="-k test ; rm -rf /"` to wipe my HD. This is an extreme example, but still you may forgot some quotes here and there and run a completely different set of tests and get a green build when it should be red or the opposite.

mkmkme:
Closed in favor of #106

PR #112, opened 2024-10-15
parsers: resp3: be less verbose

### LINKED ISSUE #111: protocole 3 (RESP3) outputs debug info with pubsub
When setting protocol=3 (RESP3), and using the pubsub features, on a message reception it seems some data is outputted in stdout by the library (or stderr not sure), here's an example of this data:
Push response: [b'subscribe', b'mychannel', 1]

This does not happen with the default protocol.

I'm using valkey-py 6.0.2

aiven-sal:
Close #111


mkmkme (approved):
LGTM Thanks!

PR #113, opened 2024-10-15
Temporarily fix https://github.com/actions/runner-images/issues/10781

aiven-sal:
Temporarily fix https://github.com/actions/runner-images/issues/10781

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -27,7 +27,7 @@ jobs:
 
    dependency-audit:
      name: Dependency audit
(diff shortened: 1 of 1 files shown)


mkmkme:
By "temporarily" you mean we should ideally support 24.04 in CI? Or did you mean something else?

mkmkme (approved):
LGTM

aiven-sal:
By "temporarily" I meant that it won't be needed after https://github.com/actions/runner-images/issues/10781  is fixed.
It looks like it was fixed, so maybe we can already revert this.

PR #125, opened 2024-11-27
Remove expiration/TTL commands that are not supported by Valkey

### LINKED ISSUE #78: hexpire exposed even though valkey doesn't support it.
**Version**: valkey-py: 6.0.0 (installed using `pip install "valkey[libvalkey]"`) valkey: docker container valkey/valkey:8.0

**Platform**: Python 3.11.6 on Manjaro Linux 

**Description**:
Should `hexpire` be exposed as a command in Valkey-py if Valkey doesn't support it?
Ideally `hexpire` is implemented, and everything is fine, but now it seems to be in a weird limbo state.

How to reproduce:

Run container: `docker run -it --rm -p 6379:6379 --name valkey valkey/valkey:8.0`

python:
```py
import valkey
r = valkey.Valkey(host='localhost', port=6379, db=0)
event = {
     'air_quality': 256,
     'battery_level':89
 }
r.hset('yolo', mapping=event)
r.hexpire('yolo', 60, 'air_quality')
>>> ResponseError: unknown command 'HEXPIRE', with args beginning with: 'yolo' '60' 'FIELDS' '1' 'air_quality' 
```

aiven-sal:
Support for this commands may come in the future[1]. But it will take some time, so for now it's better to drop them.

This is a breaking change for 6.1.

Close #78

[1]: https://github.com/valkey-io/valkey/issues/640


mkmkme (approved):
LGTM thanks!

PR #134, opened 2024-12-03
Merge original typing from types-redis, adapt it for valkey.

Rafiot:
Replacement PR for https://github.com/valkey-io/valkey-py/pull/93, this time with signoff (sorry for the noise, I had a fight with git and it seems to be the easiest way out).

Still not ready for merging.


Rafiot:
@amirreza8002 Let's discuss about the state of the PR here.

So my approach was to use the test suite as a baseline to have working type hints, so I gradually enable mypy on the files in the test suite: https://github.com/Rafiot/valkey-py/blob/types-signed-off-take2/.mypy.ini

And adapt either the test suite, or the typing, depending on the issues reported by mypy.

One day, we should try to get rid of the `.pyi` files, but it's gonna be tricky due to the way the sync and async modules are implemented.

amirreza8002:
hi
thanks for the info

and what parts do you need help with?

or what parts you aren't gonna do for now? so we don't end up working on the same thing

Rafiot:
I'm working on something else right now, so I don't have anything going that is not in this PR. Feel free to work on anything you want.

The next steps are to run mypy on more of the test files, so feel free to pick anything you want. If you have something ongoing and you don't push it same day, please add a comment here so we don't duplicate the work.

Thanks!

mkmkme:
let's instead focus on #265. Tagging this PR there keeps the record, so we might refer to it when needed. Thanks a lot one more time for all your effort!

PR #144, opened 2024-12-06
fixed problems in #143

### LINKED ISSUE #143: `Valkey.set()` checks
hi

if we pass both `ex` and `px` to [set](https://github.com/valkey-io/valkey-py/blob/main/valkey/commands/core.py#L2248), it will be passed down to valkey, which thros a syntax error, but the error is not descriptive, so maybe adding a check would be good.

also there is no type check for [exat](https://github.com/valkey-io/valkey-py/blob/main/valkey/commands/core.py#L2307) and [pxat](https://github.com/valkey-io/valkey-py/blob/main/valkey/commands/core.py#L2312) and [keepttl](https://github.com/valkey-io/valkey-py/blob/main/valkey/commands/core.py#L2317) and [nx and xx](https://github.com/valkey-io/valkey-py/blob/main/valkey/commands/core.py#L2320-L2323)

also [get](https://github.com/valkey-io/valkey-py/blob/main/valkey/commands/core.py#L2325-L2327) looks like a bug, it should be the value of get that is passed down, not hardcoded `True`

the `options` dict also seems un-needed

let me know what you think

amirreza-sf80:
closes #143

### DIFF
--- a/valkey/commands/core.py
+++ b/valkey/commands/core.py
@@ -2287,6 +2287,11 @@ class BasicKeyCommands(CommandsProtocol):
 
         For more information see https://valkey.io/commands/set
         """
+        params = sum(op is not None for op in [ex, px, exat, pxat])
+        if params > 1:
+            raise DataError(
+                "``ex``, ``px``, ``exat`` and ``pxat`` " "are mutually exclusive."
+            )
         pieces: list[EncodableT] = [name, value]
         options = {}
         if ex is not None:
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -1320,6 +1320,26 @@ class TestValkeyCommands:
         assert r.get("integer") == str(integer).encode()
         assert r.get("unicode_string").decode("utf-8") == unicode_string
 
+    def test_set_options_mutually_exclusive(self, r):
+        with pytest.raises(exceptions.DataError):
+            r.set("test", 1, ex=1, px=1)
+
+        with pytest.raises(exceptions.DataError):
+            r.set("test2", 2, exat=3, pxat=5)
+
+        with pytest.raises(exceptions.DataError):
+            r.set("test3", 3, ex=5, exat=1)
+
+    def test_set_options_type_check(self, r):
+        with pytest.raises(exceptions.DataError):
+            r.set("test", 1, ex="hi")
+
+        with pytest.raises(exceptions.DataError):
+            r.set("test1", 3, px=object())
+
+        with pytest.raises(exceptions.DataError):
+            r.set("test3", 1, pxat=3j)
+
     @skip_if_server_version_lt("6.2.0")
     def test_getdel(self, r):
         assert r.getdel("a") is None


aiven-sal:
Thanks! Could you also add test cases?

mkmkme:
Could you please add tests for both improvements you introduced?

amirreza8002:
i pushed the last commit to be discussed here

i think since the type hint is explicit that no string should be passed, having those two lines is a bit weird

mkmkme:
@aiven-sal we discussed with @amirreza8002 in matrix channel the change in [c4de8a5](https://github.com/valkey-io/valkey-py/pull/144/commits/c4de8a5d97ad3c0d080c4430f643cbca5164fc86). The current typing of the function does not allow passing a string as `ex`, so passing it at least will trigger the `mypy` warning. So this extra case looks really off, and it's quite inconsistent with the rest of the arguments. I agreed to remove it.

Is it okay with you?

mkmkme:
@amirreza8002 the linting has failed. Could you fix it please?

mkmkme:
So it seems that at least there were some users of `ex` being string :) So maybe we should not remove that handling...

amirreza8002:
the test fail is for the last commit, I'll adjust the test if the change is approved,

sorry about the lint, I'm used to my ide doing it :/
I'll fix that up

mkmkme:
So @amirreza8002 would you please revert the last commit instead of fixing the test?

We really don't want to break anything right now

amirreza8002:
> My impression (which might be not 100% right) is that many valkey-py functions accept ints and numerical strings interchangeably. This may be not always intentional, but disallowing strings now is a breaking change. I tend to be against breaking changes, but if we do it in a major release I won't oppose it.

so do you think, for a major version, other parameters should alow strings, or should we remove the string support?
e.g. right now onle `ex` allows string, `px`, `exat` and `pxat` don't

mkmkme:
>so do you think, for a major version, other parameters should alow strings, or should we remove the string support?
e.g. right now onle ex allows string, px, exat and pxat don't

I'd personally leave the arguments as they are currently. And we can put a mental note for the next major release to remove string support for `ex`

mkmkme:
Hey @amirreza8002 ! Do I understand correctly that the changes are ready for review and to be merged?

If so, would you rebase your PR on the top of the current `main` branch?

amirreza8002:
hi
yes, afaik it can be merged

aiven-sal:
> hi is this a problem from my side? the test's error doesn't seem relevant to my code I'm not sure what the other fail is for

Hi.
The problem is that you merged instead of rebasing.
We don't want merge commits in PRs (they make the git history difficult to look back at).
To fix the problem you should rebase and remove the merge commit.

mkmkme:
I think it should be fixed now, yes. I've approved the github actions.

>the problem was that i thought using github's sync would rebase, it seems like it doesn't 

AFAIK you can choose the way you want to update the branch.

valkey/commands/core.py
```diff
@@ -2311,11 +2316,15 @@ def set(
             pieces.append("EXAT")
             if isinstance(exat, datetime.datetime):
                 exat = int(exat.timestamp())
+            elif not isinstance(exat, int):
```
aiven-sal, line 2319: I'm not sure I understand this (and the other similar change below).
Shouldn't exat and pxat accept also strings? Or is there something else that prevents them from being string?

amirreza8002: I don't think it ever supported strings for exat,
am i missing something?
i can add support for strings, but my understanding of our conversation above was that i shouldn't, and the current string support for ex and px are to be removed in future

amirreza8002: i stand corrected, i'll remove the check

amirreza8002:
i have some follow up for this, but i think it should be done in another pr

mkmkme:
>i have some follow up for this, but i think it should be done in another pr

What do you have in mind?

amirreza8002:
i have three things in mind
1. if valkey accepts strings, why cast it to `int` in `ex`

2. for px, exat and pxat it might be appropriate to add check to see if the string is digit, and not `hi`

3. if string is supported might as well add it to type hints

mkmkme:
>if valkey accepts strings, why cast it to int in ex

I'm not a Valkey expert, but my understanding is that it only supports the string representations of numbers. But I'll have to double-check that.

In general, your questions are valid, but I'd hesitate to change any user experience at this moment.

mkmkme (approved):
The changes now look good to me.
Thanks! And sorry it was a bit painful to land :)

amirreza8002:
> but I'd hesitate to change any user experience at this moment.

I'm not sure how this effects user xp, but i respect the packages development plans

I was just thinking ahead, i don't think I'll be working on it any time soon, probably do some type hints for the time being since I'm doing similar work in django-valkey

PR #150, opened 2024-12-12
Fix excessive CPU usage due to polling with no sleep

NadavTasher:
### Pull Request check-list

### Description of change

Fixed excessive CPU usage by changing the default timeout values from `0` to none.

### DIFF
--- a/tests/test_asyncio/test_pubsub.py
+++ b/tests/test_asyncio/test_pubsub.py
@@ -40,7 +40,7 @@ async def wait_for_message(pubsub, timeout=0.2, ignore_subscribe_messages=False)
     timeout = now + timeout
     while now < timeout:
         message = await pubsub.get_message(
-            ignore_subscribe_messages=ignore_subscribe_messages
+            ignore_subscribe_messages=ignore_subscribe_messages, timeout=0.01
         )
         if message is not None:
             return message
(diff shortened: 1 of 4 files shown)


NadavTasher:
https://github.com/valkey-io/valkey-py/blob/f8fc0650f58f3d41ad92fb5a5ef540993b11e386/valkey/client.py#L1061
Here for example

aiven-sal:
> https://github.com/valkey-io/valkey-py/blob/f8fc0650f58f3d41ad92fb5a5ef540993b11e386/valkey/client.py#L1061
> 
> Here for example

Maybe it's because I didn't get my coffee yet, but I don't see any loop that could consume CPU. Could you point me to where that happens exactly?

NadavTasher:
My apologies – this is an issue I solved a while back for redis-py and forgot what it was about.

This issue only seems to appear when the user runs get_message() repeatedly in order to filter/receive messages in a loop.

The issue is that get_message() without any parameters is supposed to be blocking, but it is not, since the default parameter is 0.0 and not None (even though it seems like None was the intention, since it is compared to inside the function).

This solved the redis-py issue https://github.com/redis/redis-py/issues/3208, but since redis-py doesn't get much activity (and since I have switched to valkey), I thought it would be easier to fix it here.

aiven-sal:
Ah yes, thank you for taking the time to explain!

I'm not the original author of the code, but I disagree with 

> get_message() without any parameters is supposed to be blocking

I don't think that there is anything indicating that get_message without parameters should be blocking, the documentation doesn't say that. Also, the fact that your change breaks a few tests is also an indication that this behavior was intended.

I agree that many APIs default to blocking behavior and, for this reason, one might intuitively expect this to work in a similar way without reading the doc, but this is the way it was written a long time ago.
Changing it now has the potential to break other users of the library, so I don't think we can merge this PR.

The only thing we can do is to make the documentation more explicit about get_message's behavior to avoid confusion.

NadavTasher:
I understand.

I think the documentation should be clearer.
I also think that even if this PR is not mereged, we should add some warning mechanism.
I suggest we add `warnings.warn("get_message is non-blocking, 0 timeout was provided")` as a way to warn users of this dangerous behaviour.

aiven-sal:
I crated an issue to track this https://github.com/valkey-io/valkey-py/issues/151

PR #155, opened 2024-12-16
fixed type hint of hrandfield method

amirreza8002:
hrandfield returns a byte if no count is given
if count is given it returns a list of bytes
if there is no value found, it returns None


aiven-sal (dismissed):
Hi! 3.9 doesn't support this syntax. You should use Union.

amirreza8002:
ah
sorry, was doing the same thing in another package and forgot

mkmkme (approved):
LGTM, thanks!

mkmkme (dismissed):

PR #167, opened 2025-01-10
Add dynamic_startup_nodes parameter to async ValkeyCluster

Kakadus:
### Pull Request check-list

- [X] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?

### Description of change

Note: I opened this PR [some time ago at redis-py](https://github.com/redis/redis-py/pull/3447), but there seems to be no responses so far. It fixes a bug we currently have with redis-py, and we will migrate to valkey-py if this get accepted here : )

The async version of the `ValkeyCluster` lackes the `dynamic_startup_nodes` parameter. This PR adds the parameter to be activated when dynamic DNS endpoints for startup nodes are in use. 

Enabling `dynamic_startup_nodes` fixes https://github.com/redis/redis-py/issues/2472, which describes the problem in detail.

### DIFF
--- a/tests/test_asyncio/test_cluster.py
+++ b/tests/test_asyncio/test_cluster.py
@@ -2598,6 +2598,27 @@ class TestNodesManager:
                     assert rc.get_node(host=default_host, port=7001) is not None
                     assert rc.get_node(host=default_host, port=7002) is not None
 
+    @pytest.mark.parametrize("dynamic_startup_nodes", [True, False])
+    async def test_init_slots_dynamic_startup_nodes(self, dynamic_startup_nodes):
+        rc = await get_mocked_valkey_client(
+            host="my@DNS.com",
+            port=7000,
+            cluster_slots=default_cluster_slots,
+            dynamic_startup_nodes=dynamic_startup_nodes,
+        )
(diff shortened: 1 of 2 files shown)


mkmkme (approved):
Hey @Kakadus,

Thanks for your contribution! The change looks good to me. Even though it changes the default user-facing behaviour I suppose it should be fine since it shouldn't break any existing setup. I'll still wait for other maintainers to hear from them about it.

mkmkme:
Also one request: would you be able to change your `user.email` git config to an actual e-mail instead of one `@users.noreply.github.com`?

Kakadus:
> Even though it changes the default user-facing behaviour I suppose it should be fine since it shouldn't break any existing setup. I'll still wait for other maintainers to hear from them about it.

It really shouldn't? By default, `dynamic_startup_nodes` is true, so the exact same code (`self.set_nodes(...)`) should be executed. Old code without this parameter don't see any behavior change.

aiven-sal (approved):
Thank you!

mkmkme:
Great, thank you for fixing the DCO! I have put this PR into the merge queue. This change will be a part of upcoming 6.1.0 :)

PR #174, opened 2025-02-11
tests: Use fork instead of forkserver

bugant:
Make sure test can be successfully run on pyton 3.14.

According to
https://docs.python.org/dev/whatsnew/3.14.html#multiprocessing

The default start method (see Contexts and start methods) changed from fork to forkserver on platforms other than macOS & Windows where it was already spawn. If you require the threading incompatible fork start method you must explicitly request it using a context from multiprocessing.get_context() (preferred) or change the default via multiprocessing.set_start_method().

See also https://bugzilla.redhat.com/show_bug.cgi?id=2326934 for reference on the reported bug for the Fedora package.

### Pull Request check-list

### DIFF
--- a/tests/test_multiprocessing.py
+++ b/tests/test_multiprocessing.py
@@ -29,7 +29,14 @@ class TestMultiprocessing:
             valkey.Valkey, request=request, single_connection_client=False
         )
 
-    def test_close_connection_in_child(self, master_host):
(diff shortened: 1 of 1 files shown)


mkmkme (approved):
LGTM! Thanks for the swift fix. It will be part of 6.1.1

PR #170, opened 2025-01-31
Update integration.yaml 

NarmalaSk:
### Pull Request check-list

- [x] Do tests and lints pass with this change?
- [ ] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)?
- [x] Is the new or changed code fully tested?
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [x] Is there an example added to the examples folder (if applicable)?

PR #176, opened 2025-02-26
github: fix caching the docker images

### LINKED ISSUE #140: docker cache not updated
The docker cache used by the CI runs is supposed to be updated every night.
But the nightly update doesn't work. The only way to update it is manually clear the cache.
We should probably fix [this if](https://github.com/valkey-io/valkey-py/blob/main/.github/workflows/integration.yaml#L65) but I'm not 100% sure.

mkmkme:
This PR supercedes #170 and fixes #140 

### Pull Request check-list

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -50,8 +50,19 @@ jobs:
            pip install -r dev_requirements.txt
            invoke linters
 
+   get-date:
+    name: Get current date for cache
+    outputs:
+      current_date: ${{ steps.date.outputs.current_date }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set current date
+        id: date
+        run: echo "current_date=$(date -u +'%Y%m%d')" >> $GITHUB_OUTPUT
+
    populate-cache:
      runs-on: ubuntu-latest
+     needs: [get-date, dependency-audit, lint]
      timeout-minutes: 60
      name: Update docker cache
      steps:
@@ -61,8 +72,8 @@ jobs:
          uses: actions/cache@v4
          with:
            path: ./custom-cache/
-           key: custom-cache
-       - if: ${{ steps.custom-cache.outputs.cache-hit != 'true' || github.event_name == 'schedule' }}
+           key: custom-cache-${{ needs.get-date.outputs.current_date }}
+       - if: ${{ steps.custom-cache.outputs.cache-hit != 'true' }}
          name: Update Cache
          run: |
            mkdir -p ./custom-cache/
@@ -72,7 +83,7 @@ jobs:
 
    run-tests:
      runs-on: ubuntu-latest
-     needs: [populate-cache]
+     needs: [populate-cache, get-date]
      timeout-minutes: 60
      strategy:
        max-parallel: 15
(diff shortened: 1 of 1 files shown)


mkmkme:
Expanded the commit message a bit:

> This commit updates the way docker cache is created by including the
current date into the cache key. Now when "save-always" option in
actions/cache is deprecated [1] this seems to be the conventional
way [2] to create the cache.

1: https://github.com/actions/cache/pull/1452
2. https://github.com/actions/cache?tab=readme-ov-file#creating-a-cache-key

Also removed the `event_name == 'schedule'` check since it doesn't make much sense with the current approach anymore. The cache will be created every day (like it should have been with the original approach). And the old cache will be removed after 7 days of not accessing it according to Github policy: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy

I am especially interested in seeing @smeso review on this PR because you created the original caching approach.

.github/workflows/integration.yaml
```diff
@@ -98,7 +104,7 @@ jobs:
          with:
            path: ./custom-cache/
            fail-on-cache-miss: true
-           key: custom-cache
+           key: custom-cache-${{ steps.date.outputs.current_date }}
```
bogdanp05, line 107: Is it possible the `current_date` computed here is different from the `current_date` computed in `populate-cache` if the pipeline is run around midnight?

mkmkme: True, this is possible, and the pipeline might fail because of that

mkmkme: It seems like we can't possibly set a *global* variable of a date. So the only possible way to fix this is to repeat the same "look for cache -> if not found, create one" pattern here as well just because of this corner case and Github limitations :/ That's unfortunate.

mkmkme: There are also other options, like:
1. Create cache for each PR instead of tying it with the date
2. Just accept the occasional PR github action failures and restart them manually. It shouldn't happen that often anyways.

bogdanp05: I thought it would be possible to create a job to get the date and then just add a dependency to it, at least according to [the docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/passing-information-between-jobs). So something like:
```
   get-date:
     outputs:
       current_date: ${{ steps.date.outputs.current_date }}
     steps:
       - name: Set current date
         id: date
         run: echo "current_date=$(date +'%Y%m%d')" >> $GITHUB_OUTPUT
```
and then use it like:
```
   run-tests:
     runs-on: ubuntu-latest
     needs: [populate-cache, get-date]
     ....
        with:
           path: ./custom-cache/
           fail-on-cache-miss: true
           key: custom-cache-${{ needs.get-date.outputs.current_date }}
```
I haven't tested this, but if it works, then it would be nice because it abstracts the duplicate "Set current date" step.

If it doesn't work, then I'm happy with option 2. That race condition would happen super rarely anyway and it would work on the next try.

mkmkme: Hmmm, thanks, I'll try!

mkmkme:
Hey @bogdanp05 , I've fixed our conversation. Your approach worked, thanks! Please approve if it looks good to you, it'll auto-merge.

ahmedsobeh (approved):

PR #194, opened 2025-05-12
tests: use `sorted(...)` instead of `.sort()`

Kakadus:
### Pull Request check-list

### Description of change

While doing #167, I just copied existing tests for the valkey cluster. After reading the test code again, I noticed that the code compares 
```py
assert some_list.sort() == other_list.sort()
```

This is inherently flawed, as `.sort()` returns `None`. You can verify that with `[0,1].sort() == [1,2].sort()`. 

This PR uses `sorted(...)` instead which actually returns the sorted list as intended. I also fix one other occurrence of this I found using `rg "sort\(\) =="`


bogdanp05 (approved):
nice catch!
lgtm

PR #212, opened 2025-07-09
Documentation fixes

### LINKED ISSUE #133: Add reference to Valkey's matrix channel in doc
We should refer people to matrix for generic discussions or help question

### LINKED ISSUE #191: cluster scan returns a dict as cursor
hi
so the cluster client runs this function on the return value of scan https://github.com/valkey-io/valkey-py/blob/main/valkey/cluster.py#L56

which would look something like this in a test example:
```
cursor={'127.0.0.1:7000': 0, '127.0.0.1:7001': 0, '127.0.0.1:7002': 0}
result=['foo2', 'foo3', 'foo1']
```

i understand that this is intended behaviour, but am wondering if this is necessary
valkey itself returns a single number

but if it's here to stay, we should probably document it

mkmkme:
### Pull Request check-list

- [X] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?

### Description of change

Some minor documentation changes needed before releasing 6.1.1. Fixes #133, fixes #191

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -9,7 +9,7 @@ valkey-py - Python Client for Valkey
 Getting Started
 ****************
 
-`valkey-py <https://pypi.org/project/valkey>`_ requires a running Valkey server, and Python 3.7+. See the `Valkey
+`valkey-py <https://pypi.org/project/valkey>`_ requires a running Valkey server, and Python 3.9+. See the `Valkey
 quickstart <https://valkey.io/topics/quickstart>`_ for Valkey installation instructions.
 
 valkey-py can be installed using pip via ``pip install valkey``.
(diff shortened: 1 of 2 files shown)


mkmkme:
@amirreza8002 could you have a look at this as the author of #191  ?

amirreza-sf80:
hi
i was thinking, maybe the return type should be adjusted as well

i think it'd be something like:
`Union[tuple[int, list[KeyT]], Awaitable[tuple[int, list[KeyT]]], tuple[dict[str, int], list[KeyT]], Awaitable[tuple[dict[str, int], list[KeyT]]`

mkmkme:
@amirreza8002 let's not fix the types in the scope of this PR. Typing is a much larger problem that doesn't have to be fixed by 6.1.1. Documentation OTOH is more important. Current return type is not incorrect, it's just too wide.

bogdanp05 (approved):

PR #230, opened 2025-10-07
Switch from setup.py to pyproject.toml

mkmkme:
Highly inspired by #96 by @ds-cbo (Thanks!)

### Pull Request check-list

### Description of change

Switch from setup.py to pyproject.toml


mkmkme:
@trim21 would you be able to review this one? Thanks!

pyproject.toml
```diff
@@ -0,0 +1,96 @@
+[build-system]
+requires = ["setuptools>=61.0", "wheel"]
```
trim21, line 2: wheel is not necessary anymore

pyproject.toml
```diff
@@ -0,0 +1,96 @@
...
+    "valkey.commands",
+    "valkey.commands.bf",
+    "valkey.commands.json",
+    "valkey.commands.search",
+    "valkey.commands.timeseries",
+    "valkey.commands.graph",
+]
+
+[tool.setuptools.package-data]
+valkey = ["py.typed"]
```
trim21, line 96: This is also not necessary

bogdanp05 (dismissed):

ahmedsobeh (dismissed):

mkmkme:
Thanks everyone for the reviews! I'll double-check tonight whether the changes in the comments are really needed and will either fix or merge this

pyproject.toml
```diff
@@ -0,0 +1,96 @@
...
+requires-python = ">=3.10"
+
+dependencies = [
+    'async-timeout>=4.0.3; python_full_version<"3.11.3"',
+]
+
+[project.optional-dependencies]
+libvalkey = ["libvalkey>=4.0.1"]
+ocsp = ["cryptography>=36.0.1", "pyopenssl==23.2.1", "requests>=2.31.0"]
+dev = [
```
amirreza8002, line 44: just an fyi

i think? people can now do `pip install valkey[dev]` and get from pypi
same thing goes for `doc`

some package managers (uv, poetry) have a way to prevent this

amirreza-sf80:
also you might consider a build tool that's not setuptools ;)

it lacks some features, I've had cases where it couldn't even publish to pypi cause it lacked the new standard

pyproject.toml
mkmkme: Yeah that's a great point.

If there's a way to do that with `pyproject` itself, I'm all for it. 

Alternatively, we can also do what redis-py has done, that is switching to `pyproject` but keeping `dev_requirements.txt` and `docs/requirements.txt`. Honestly this sounds like the best approach.

@bogdanp05 WDYT?

mkmkme:
> also you might consider a build tool that's not setuptools ;)
> 
> it lacks some features, I've had cases where it couldn't even publish to pypi cause it lacked the new standard

Thanks a lot! I'll look into it. Do you have any suggestions for the alternative?

trim21:
> > also you might consider a build tool that's not setuptools ;)
> > it lacks some features, I've had cases where it couldn't even publish to pypi cause it lacked the new standard
> 
> Thanks a lot! I'll look into it. Do you have any suggestions for the alternative?

flit-core

pyproject.toml
amirreza8002: https://github.com/django-commons/django-valkey/blob/main/pyproject.toml#L60

this clause is documented in pyproject docs

but if i understand correctly it needs a lock file to work, so a package manager is needed

mkmkme:
> > also you might consider a build tool that's not setuptools ;)
> > it lacks some features, I've had cases where it couldn't even publish to pypi cause it lacked the new standard
> 
> Thanks a lot! I'll look into it. Do you have any suggestions for the alternative?

Note to myself: redis-py uses `hatchling`, might worth looking into it as well.

pyproject.toml
amirreza8002: there is a chance my data is outdated and it can be used as is 
https://packaging.python.org/en/latest/specifications/dependency-groups/

amirreza-sf80:
i use `hatch` everywhere 

it does the job without noise

but I'm not an expert in this matter and don't have a strong suggestion

mkmkme:
We've decided to continue with hatchling. It seems to be more widely adopted. The commit related to that has been added to this PR.

amirreza-sf80:
did `dependency-groups` not work here?
i did a test and `pip install --group dev` works for me if `dependency-groups` is used in pyproject

mkmkme:
> did `dependency-groups` not work here? i did a test and `pip install --group dev` works for me if `dependency-groups` is used in pyproject

Hey.

Thanks for highlighting this. For some reason I thought since it was a recent addition, it would be unavailable on the older Python versions. I was wrong. I'll add it and will continue on this PR tomorrow.

mkmkme:
> did `dependency-groups` not work here? i did a test and `pip install --group dev` works for me if `dependency-groups` is used in pyproject

Added it in follow-up commits. Thanks for the hint!

bogdanp05 (approved):
nice job!

PR #106, opened 2024-10-01
replace pyinvoke with tox

mkmkme:
### Pull Request check-list

- [X] Do tests and lints pass with this change?
- [ ] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)?
- [ ] Is the new or changed code fully tested?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [ ] Is there an example added to the examples folder (if applicable)?

### Description of change

This change replaces `invoke` with `tox`

PR #54, opened 2024-07-12
Add support for Python 3.12 and enable it in CI

mkmkme:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [X] Do tests and lints pass with this change?
- [ ] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)?
- [ ] Is the new or changed code fully tested?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [ ] Is there an example added to the examples folder (if applicable)?

_NOTE: these things are not required to open a PR and can be done
afterwards / while the PR is open._

### Description of change

This changeset finishes adding support for Python 3.12. This should close #24

PR #233, opened 2025-10-18
Replace pyinvoke with Makefile

### LINKED ISSUE #62: allow for `TESTARGS` in invoke test

mkmkme:
pyinvoke has several serious issues. First and foremost, it is very unsafe to pass any string arguments to `run()`, because it does not verify the string arguments at all. Given this function in `tasks.py`:

```python
@task
def hello(c, name="user"):
    run(f"echo {name}")
```

you can get quite interesting results:

```sh
❯ invoke hello
user
❯ invoke hello --name="; date"

Sat Oct 18 11:41:56 CEST 2025

```

This is an issue number 2 in pyinvoke and I can't see it closed in any foreseeable future:
https://github.com/pyinvoke/invoke/issues/2

Additionally, running linters from `invoke.run()` disables any colors from the linters which makes it less comfortable to use and caused us to force colors in some targets:
https://github.com/valkey-io/valkey-py/pull/54/commits/6111059d1b3714e0e2910400717c5a735e591c81

Even with that change, I have to pass additional `-c` any time I would like to run linters with colored output.

Considering these main issues, I have considered different alternatives to pyinvoke with one of them being `tox`:
https://github.com/valkey-io/valkey-py/pull/106

However, `tox` seems to be somewhat of an overkill for such a simple task, and also packaging was done via `invoke` as well. Therefore I've decided to pick `make` as an alternative.

This commit replaces `tasks.py` with a corresponding `Makefile`, removes `invoke` from `dev` dependencies (additionally removing `click` that was mentioned there explicitly for some reason) and replaces all the calls for `invoke` with calls for `make`.

This PR closes #62 

### Pull Request check-list


bogdanp05 (approved):
This looks way cleaner than before, thanks!

PR #240, opened 2025-11-04
Suggestion: Add max_tries to transaction()

haukex:
Hi, I would like to propose that the `transaction()` helper gets an additional argument `max_tries` (perhaps even with a default value) to prevent infinite loops, as can happen fairly easily by accident:
```python
vk = Valkey()
def incr(pipe: Pipeline):
    next_value = int(pipe.get('key')) + 1
    pipe.multi()
    vk.set('key', next_value)  # WRONG!
vk.transaction(incr, 'key', max_tries=100)  # would eventually raise error instead of infinite loop
```
If you like this change, I can also suggest some documentation updates etc. if you like.

### DIFF
--- a/valkey/asyncio/client.py
+++ b/valkey/asyncio/client.py
@@ -448,14 +448,14 @@ class Valkey(
             `WatchError` before the transaction is retried. Default is no delay.
         :param max_tries: Lets you specify the maximum number of times the
             transaction is retried. If the limit is reached, a `ValkeyError`
-            is raised. Default is an **infinite** number of retries!
+            is raised. Default is 0, meaning an **infinite** number of retries!
         """
         pipe: Pipeline
         async with self.pipeline(True, shard_hint) as pipe:
             tries = 0
             while True:
                 tries += 1
-                if max_tries and max_tries > 0 and tries > max_tries:
+                if max_tries and tries > max_tries:
                     raise ValkeyError(
                         f"Bailing out of transaction after {tries - 1} tries"
                     )
--- a/valkey/client.py
+++ b/valkey/client.py
@@ -410,8 +410,8 @@ class Valkey(ValkeyModuleCommands, CoreCommands, SentinelCommands):
             is retried. Default is no delay.
         :param max_tries: This keyword-only argument lets you specify the
             maximum number to times the transaction should be retried. If the
-            limit is reached, a `ValkeyError` is raised. Default is an
-            **infinite** number of retries!
+            limit is reached, a `ValkeyError` is raised. Default is 0, meaning
+            an **infinite** number of retries!
         """
         shard_hint = kwargs.pop("shard_hint", None)
         value_from_callable = kwargs.pop("value_from_callable", False)
@@ -421,7 +421,7 @@ class Valkey(ValkeyModuleCommands, CoreCommands, SentinelCommands):
             tries = 0
             while True:
                 tries += 1
-                if max_tries and max_tries > 0 and tries > max_tries:
+                if max_tries and tries > max_tries:
                     raise ValkeyError(
                         f"Bailing out of transaction after {tries - 1} tries"
                     )
--- a/tests/test_asyncio/test_pipeline.py
+++ b/tests/test_asyncio/test_pipeline.py
@@ -343,15 +343,32 @@ class TestPipeline:
             pipe.multi()
             await r.set("a", a_value + 1)  # force WatchError
 
+        # without max_tries (infinite loop)
         with pytest.raises(RuntimeError) as ex:
             await r.transaction(my_transaction, "a")
         assert str(ex.value).startswith("Run too many times")
         assert run_count == 11
+
         run_count = 0
+        # with max_tries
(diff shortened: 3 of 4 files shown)


mkmkme:
Hey,

Thanks for your contribution! Could you please sign off your commit? You can find how to do it [here](https://github.com/valkey-io/valkey-py/pull/240/checks?check_run_id=54458554374), in the failed check report.

mkmkme:
Also I would really appreciate if you could provide a test for your changes

haukex:
@mkmkme Thanks, I've amended and force pushed the commit to include documentation and tests. *Edit:* Also rebased. *Edit 2:* And fixed `Signed-off-by`. *Edit 3:* And a final update with a minor improvement to the tests. Done editing for now 😄

valkey/asyncio/client.py
```diff
@@ -437,15 +437,28 @@ async def transaction(
...
         value_from_callable: bool = False,
         watch_delay: Optional[float] = None,
+        max_tries: Optional[int] = None,
     ):
         """
         Convenience method for executing the callable `func` as a transaction
         while watching all keys specified in `watches`. The 'func' callable
         should expect a single argument which is a Pipeline object.
+
+        :param watch_delay: Lets you specify a delay time in seconds after a
```
mkmkme: Kudos for adding a documentation for `watch_delay` as well!

valkey/client.py
```diff
@@ -404,12 +404,27 @@ def transaction(
...
         """
         shard_hint = kwargs.pop("shard_hint", None)
         value_from_callable = kwargs.pop("value_from_callable", False)
         watch_delay = kwargs.pop("watch_delay", None)
+        max_tries = kwargs.pop("max_tries", None)
         with self.pipeline(True, shard_hint) as pipe:
+            tries = 0
             while True:
+                tries += 1
+                if max_tries and max_tries > 0 and tries > max_tries:
```
amirreza8002:
```suggestion
                tries += 1
                if max_tries and tries > max_tries:
```

0 is falsy and doesn't need separate check

valkey/asyncio/client.py
```diff
@@ -437,15 +437,28 @@ async def transaction(
...
+        :param max_tries: Lets you specify the maximum number of times the
+            transaction is retried. If the limit is reached, a `ValkeyError`
+            is raised. Default is an **infinite** number of retries!
         """
         pipe: Pipeline
         async with self.pipeline(True, shard_hint) as pipe:
+            tries = 0
             while True:
+                tries += 1
+                if max_tries and max_tries > 0 and tries > max_tries:
```
amirreza8002:
```suggestion
                if max_tries and tries > max_tries:
```
0 is falsy

amirreza-sf80:
i haven't worked with this method, but it looks ok, tho added a comment

valkey/client.py
mkmkme: This particular change would make it so negative value could be passed to an unexpected behaviour. Maybe something like this?

```python
if max_tries is not None and max_tries > 0 and tries > max_tries
```

I would also expect `max_tries=0` work as "repeat infinitely"

amirreza8002: what would be the bahaviour if negetive is passed?

mkmkme: Ah, my bad, I misread the code. I thought it would go from negative number until it reaches 0. But actually it would loop indefinitely, which would be something I would expect here as well. So I guess it's fine :)

haukex:
After thinking about this a bit more, I think I actually prefer the updated version `if max_tries and tries > max_tries:` that raises an error immediately if a negative value is passed. This seems more logical to me since a negative number of tries doesn't make sense - and could even be considered worth raising a `ValueError`. (Plus, it leaves the door open for perhaps assigning negative numbers a different meaning in the future instead of immediately assigning it the meaning "an infinite number of tries".) But this is still just my opinion and I'll defer to what you prefer.

valkey/client.py
mkmkme: I meant applying Amirreza's, yes

haukex: Since I agree with the proposal to change `if max_tries and max_tries > 0 and tries > max_tries:` to `if max_tries and tries > max_tries:` (causing negative `max_tries` to be an immediate error), and my second commit does that, so I'll mark this as resolved for now.

haukex:
This latest force push was just a rebase. *Edit:* Sorry, that seems to have messed up the DCO, that should now be fixed.

haukex:
@mkmkme @amirreza8002 @bogdanp05 Happy new year! and Bump. From my side I think this PR is ready to go.

amirreza-sf80:
hi, happy new year
i'm not a maintainer here, i can't do anything in regards to this
i just look around and throw some random statements at PRs :p

bogdanp05:
@haukex Happy new year! Thanks for the contribution and for the follow up. I approved the PR, could you rebase pls?

haukex:
@bogdanp05 Thanks very much, done!

Issue #242, opened 2025-11-05
UTF-8 Encoding Issue That Corrupts Binary Vector Data in Search Results


swarnaprakash:
## Summary

The `Result` class in `valkey/commands/search/result.py` inappropriately applies UTF-8 decoding to all field values, including binary vector data. This corrupts VECTOR field embeddings and makes valkey-py unsuitable for vector search applications.

## Root Cause

In `valkey/commands/search/result.py` line ~45:
```python
map(to_string, res[i + fields_offset][1::2])
```

The `to_string()` function uses `decode("utf-8", "ignore")` on all field values. VECTOR fields contain binary FLOAT32 data with invalid UTF-8 bytes, and the `"ignore"` parameter silently drops bytes, corrupting the data.

## Reproduction 

Run  https://gist.github.com/swarnaprakash/0ae1e454cfae3ab2a0f628181d4e8045  to see the issue:

```bash
Valkey-py Vector Search Corruption Test
=============================================
Connected to Valkey server
Created vector search index with 2 vector fields

Inserting test documents with vector embeddings:
  doc:1: embedding1=[0.1, 0.2, 0.3, 0.4], embedding2=[0.5, 0.6, 0.7]
  doc:2: embedding1=[0.8, 0.9, 1.0, 1.1], embedding2=[1.2, 1.3, 1.4]
  doc:3: embedding1=[1.5, 1.6, 1.7, 1.8], embedding2=[1.9, 2.0, 2.1]

Performing vector search with query: [0.11, 0.21, 0.31, 0.41]
Found 3 documents

==================================================
TESTING VECTOR RECONSTRUCTION
==================================================

Document 1: doc:1
  embedding1 field type: <class 'str'>
    [FAIL] embedding1 is string, not bytes: '=L>>>'
    [FAIL] Cannot reconstruct vector: unpack requires a buffer of 16 bytes
    [FAIL] Original bytes: 16 bytes
    [FAIL] Corrupted string: 5 chars
    [FAIL] DATA CORRUPTION DETECTED
    [FAIL] embedding2 corrupted: <class 'str'>

Document 2: doc:2
  embedding1 field type: <class 'str'>
    [FAIL] embedding1 is string, not bytes: 'L?fff?\x00\x00?̌?'
    [FAIL] Cannot reconstruct vector: unpack requires a buffer of 16 bytes
    [FAIL] Original bytes: 16 bytes
    [FAIL] Corrupted string: 11 chars
[...]

PR #243, opened 2025-11-05
Fix/result embedding data encoding

swarnaprakash:
### Pull Request check-list

- [Y] Do tests and lints pass with this change? _Ran linter and corrected changed lines_
- [Y ] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)? See https://github.com/swarnaprakash/valkey-py/actions/runs/19217879094 . The failed tests are unrelated to this change
- [Y] Is the new or changed code fully tested? _added tests for the new methods and enabled it (even though existing tests don't pass)_
- [Y] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? T_he docstring is updated withe new parameters_
- [Y] Is there an example added to the examples folder (if applicable)? _NA_

### Description of change

See https://github.com/valkey-io/valkey-py/issues/242

The `Result` class in `valkey/commands/search/result.py` inappropriately applies UTF-8 decoding to all field values, including binary vector data. This corrupts VECTOR field embeddings and makes valkey-py unsuitable for vector search applications.

The change  Adds preserve_bytes and binary_fields parameters to search methods to prevent UTF-8 decoding from corrupting VECTOR field embeddings and other binary data.
 The Result class was inappropriately applying UTF-8 decoding to all field values, including binary vector embeddings. This corrupted FLOAT32 vector data and made     valkey-py unsuitable for vector search applications.

### DIFF
--- a/tests/test_search_binary_preservation.py
+++ b/tests/test_search_binary_preservation.py
@@ -1,0 +1,192 @@
+"""
+Binary preservation tests for search results.
+
+These tests are in a separate file because the main search test suite (test_search.py)
+has compatibility issues with the current Valkey search module version. Most existing
+search tests fail due to unsupported field types and parameters (e.g., TEXT fields,
+SKIPINITIALSCAN, etc.).
+
+Our binary preservation functionality works correctly with the current search module
+using direct FT.CREATE commands and KNN vector queries, so we maintain these tests
+separately to ensure the feature remains properly tested while the broader search
+test compatibility issues are resolved.
+"""
+
+import struct
+
+import pytest
+import valkey
+
+from .conftest import _get_client, is_resp2_connection, skip_ifmodversion_lt
+
+
+@pytest.mark.valkeymod
+@skip_ifmodversion_lt("1.0.0", "search")
+def test_vector_binary_preservation_default_behavior(request):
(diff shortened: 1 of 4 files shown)


mkmkme:
@amirreza8002 is this by any chance something you got to work with?
@bogdanp05 I understand it's very unlikely you have worked with this before, but could you also have a look when you have time please?

mkmkme:
Also keep in mind approving this PR is slightly more problematic due to the fact valkey-search is not yet supported in the CI and the pytest skips that. I'll try to see if it can be easily fixed.

swarnaprakash:
> Also keep in mind approving this PR is slightly more problematic due to the fact valkey-search is not yet supported in the CI and the pytest skips that. I'll try to see if it can be easily fixed.

@mkmkme I can make an attempt to fix the tests. First I want to clarify the expectation.  Is valkey-py expected to be backward compatible with redis? specifically w.r.t search module is the expectation still backward compatibility with Redis search. I understand redis supports different data types (TEXT instead of TAG for example).  Wondering if I can just delete/modify these or I need to keep them and make changes such that both redis search and valkey-search are supported

amirreza-sf80:
sorry not in my area
i might take a look if i find the time

PR #246, opened 2025-11-07
Add HSETEX

### LINKED ISSUE #222: Add support for HSETEX
https://valkey.io/commands/hsetex/

jogehl:
### Pull Request check-list

- Fix #222 

Adds tests and the command HSETEX

### DIFF
--- a/valkey/commands/core.py
+++ b/valkey/commands/core.py
@@ -5083,6 +5083,10 @@ class HashCommands(CommandsProtocol):
 
         if key is None and not mapping and not items:
             raise DataError("'hsetex' with no key value pairs")
+        if int(key is not None) + int(mapping is not None) + int(items is not None) > 1:
+            raise DataError(
+                "Only one of 'key/value', 'mapping', or 'items' can be specified."
+            )
 
         if int(keepttl) + sum(arg is not None for arg in [ex, px, exat, pxat]) > 1:
             raise DataError(
--- a/valkey/commands/core.py
+++ b/valkey/commands/core.py
@@ -5076,6 +5076,8 @@ class HashCommands(CommandsProtocol):
         ``mapping`` accepts a dict of key/value pairs to be added to hash ``name``.
         ``items`` accepts a list of key/value pairs to be added to hash ``name``.
         Set expiration options for the hash fields.
+
+        For more information see https://valkey.io/commands/hsetex
         """
 
         if key is None and not mapping and not items:
--- a/valkey/commands/core.py
+++ b/valkey/commands/core.py
@@ -5066,6 +5066,7 @@ class HashCommands(CommandsProtocol):
         px: Union[ExpiryT, None] = None,
         exat: Union[AbsExpiryT, None] = None,
         pxat: Union[AbsExpiryT, None] = None,
+        keepttl: bool = False,
         nx: bool = False,
         xx: bool = False,
         fnx: bool = False,
@@ -5082,6 +5083,15 @@ class HashCommands(CommandsProtocol):
 
         if key is None and not mapping and not items:
             raise DataError("'hsetex' with no key value pairs")
+
+        if int(keepttl) + sum(arg is not None for arg in [ex, px, exat, pxat]) > 1:
+            raise DataError(
+                "Only one of 'ex', 'px', 'exat', 'pxat', or 'keepttl' can be specified."
+            )
+        if nx and xx:
+            raise DataError("Only one of 'nx' or 'xx' can be specified.")
+        if fnx and fxx:
+            raise DataError("Only one of 'fnx' or 'fxx' can be specified.")
         pieces = []
         if ex is not None:
             pieces.extend(["EX", ex])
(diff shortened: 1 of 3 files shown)


mkmkme:
Hey @jogehl,

Thanks a lot for your contribution, I appreciate that! Could you please force-push your commit with a signoff? We require all of the commits to be signed, so currently the DCO check fails on your PR. You can find the instructions of how to do that [here](https://github.com/valkey-io/valkey-py/pull/246/checks?check_run_id=54832609856)

valkey/commands/core.py
```diff
@@ -5055,6 +5055,66 @@ def hsetnx(self, name: str, key: str, value: str) -> Union[Awaitable[bool], bool
...
+        if key is None and not mapping and not items:
+            raise DataError("'hsetex' with no key value pairs")
+        pieces = []
+        if ex is not None:
+            pieces.extend(["EX", ex])
+        if px is not None:
+            pieces.extend(["PX", px])
+        if exat is not None:
+            pieces.extend(["EXAT", exat])
+        if pxat is not None:
```
mkmkme: I think it would be better to provide the check for only one of `EX`, `PX`, `EXAT`, `PXAT` being provided.
Also IMHO it would make sense to add `KEEPTTL` as well.

valkey/commands/core.py
```diff
@@ -5055,6 +5055,66 @@ def hsetnx(self, name: str, key: str, value: str) -> Union[Awaitable[bool], bool
...
+            pieces.extend(["PXAT", pxat])
+        if nx:
+            pieces.append("NX")
+        if xx:
```
mkmkme: Also check for only one of `NX`, `XX` being provided would be great

valkey/commands/core.py
```diff
@@ -5055,6 +5055,66 @@ def hsetnx(self, name: str, key: str, value: str) -> Union[Awaitable[bool], bool
...
+            pieces.append("XX")
+        if fnx:
+            pieces.append("FNX")
+        if fxx:
+            pieces.append("FXX")
```
mkmkme: Same with `FNX` and `FXX`

valkey/commands/core.py
```diff
@@ -5055,6 +5055,66 @@ def hsetnx(self, name: str, key: str, value: str) -> Union[Awaitable[bool], bool
...
+        if key is not None and value is not None:
+            pieces.append(1)  # for one field
+            pieces.append(key)
+            pieces.append(value)
+        if mapping:
+            pieces.append(len(mapping))
+            for key, value in mapping.items():
+                pieces.append(key)
+                pieces.append(value)
+        if items:
```
mkmkme: Probably checking for only one of `key + value`, `mapping` or `items` would be great

mkmkme (dismissed):
Thanks for your contribution! Overall it looks great. I added some comments to improve it even further. Could you address them?

valkey/commands/core.py
```diff
@@ -5055,6 +5055,83 @@ def hsetnx(self, name: str, key: str, value: str) -> Union[Awaitable[bool], bool
...
+
+        if key is None and not mapping and not items:
+            raise DataError("'hsetex' with no key value pairs")
+
+        def _check_for_multiple(*args):
+            return sum(arg is not None for arg in args)
+
+        if _check_for_multiple(ex, px, exat, pxat) > 1 or (
+            keepttl and _check_for_multiple(ex, px, exat, pxat) > 0
+        ):
```
mkmkme: Maybe we could simplify it to

```python
if int(keepttl) + sum(arg is not None for arg in [ex, px, exat, pxat]) > 1:
```
? We won't even need a helper function in that case.

mkmkme:
Also I've just now realised you didn't provide an async version. Could you add it and the test for it as well? It will be mostly copy-paste

jogehl:
> Also I've just now realised you didn't provide an async version. Could you add it and the test for it as well? It will be mostly copy-paste

where can I find the implementations for the async? I can't find the async version for the HashCommands.

mkmkme:
> > Also I've just now realised you didn't provide an async version. Could you add it and the test for it as well? It will be mostly copy-paste
> 
> where can I find the implementations for the async? I can't find the async version for the HashCommands.

AFAICS AsyncHashCommands is just set to be the same as HashCommands. There is a subtle difference, though. `valkey-py` has a tricky class hierarchy that effectively makes `HashCommands` use synchronous version of `execute_command` whilst `AsyncHashCommands` is using the async one.

**TL;DR:** could you add an async _test_ to verify it works? Thanks!

jogehl:
Is there anything else which is needed for this PR?

bogdanp05:
Hi @jogehl ! Looks good there's just one comment from @mkmkme that should be addressed before merging: https://github.com/valkey-io/valkey-py/pull/246#discussion_r2506388969 . Atm, the function would allow you to send something like the below to Valkey if you set more than one of (`key + value`, `mapping`, `items`).
```
127.0.0.1:6379> HSETEX myhash FIELDS 1 f4 v4 2 f5 v5 f6 v6
(error) ERR numfields should be greater than 0 and match the provided number of fields
```

jogehl:
This should be fixed now. Sorry for the delay.

bogdanp05 (approved):

bogdanp05:
merged, thanks again for the work @jogehl !

PR #265, opened 2026-02-15
Proper attempt of fixing typing warnings for valkey-py users

### LINKED ISSUE #164: Rework typing in the repo
As was pointed in #162, the typing problems go quite deep and will likely require some refactoring.

The biggest issue is the fact that sync and async clients inherit from the same class, which causes issues in type system.

For instance, this code:

```python
from typing import *

class Parent:
    def set(self) -> int:
        return 1
        
class Child(Parent):
    async def set(self) -> int:
        return 1
```

raises this mypy error:
```
main.py:8: error: Return type "Coroutine[Any, Any, int]" of "set" incompatible with return type "int" in supertype "Parent"  [override]
```

Attempting to have a union there won't help either. That is, this code

```python
from typing import *

class Parent:
    def set(self) -> int | Coroutine[Any, Any, int]:
        return 1
        
class Child(Parent):
    async def set(self) -> int | Coroutine[Any, Any, int]:
        return 1
```

causes this error

```
main.py:8: error: Return type "Coroutine[Any, Any, int | Coroutine[Any, Any, int]]" of "set" incompatible with return type "int | Coroutine[Any, Any, int]" in supertype "Parent"  [override]
```

There should be a way to fix this without completely reworking the class hierarchy in the repo. This should be investigated.

mkmkme:
### Pull Request check-list

### Description of change

This should finally fix #164.

#### Personal note

I spent quite some time on this. Initially I went through each command one by one and double-checked the following:
1. Valkey official documentation
2. A callback that parses the response in valkey-py codebase
3. Typing of redis-py in typesched repository (now deleted)
4. Actually running Python interpreter and calling the function to see what it returns

#### What is the biggest obstacle for type checkers?

As some already know, the structure of valkey-py codebase is quite neat from the implementation perspective while it's horrible for the type checkers. The main problem about it is the fact that we want to be able to call any command directly from the client:

```python
import valkey
c = valkey.Valkey()
c.lolwut()
```
... while not necessarily keeping all of the commands' implementations in the client class. We don't want that for two reasons:
1. It would make client class ENORMOUSLY large
2. We would need to do that TWO times because we have at least two main clients: a synchronous one and an asynchronous one

Therefore, the commands' implementations is split into different classes, called `*Commands`. Each of the classes is basically doing the argument parsing and passing to one core function called `execute_command`. The `execute_command` function is the one doing all the heavy lifting and all the sync or async I/O, depending on the client. Then, _the client itself_ inherits all the `*Commands` classes in order to enable the user to call any command directly from the class.

Note that in the implementation, there's always an `AsyncFooCommands` for each `FooCommands`, even though typically it is only an alias. This is an important bit for the explanation below.

#### What have I done differently now?

Our most common type checker problem is the fact that any command can be either synchronous or asynchronous, and there's no proper way for the linter to tell which one is applicable to which client. We still don't want to have two different implementations. **BUT** we can have two different *declarations*, one being sync and another being async. Like this:

```python

# Stub
class FooCommands:
    def foo(self) -> int: ...

class AsyncFooCommands:
    async def foo(self) -> int: ...

# Implementation
class FooCommands:
    def foo(self): # -> int | Awaitable[int]
        return self.execute_command("FOO")

AsyncFooCommands = FooCommands
```

This pattern seems to actually fix the static analyzers while not changing pretty much anything about the implementation. **The stubs also appear to be truthful**, because ultimately each function's `async`-ness depends on `execute_command` implementation.

Please also be aware that at some point I've given up doing everything by hand since it was taking a lot of time, and some of the functions were offloaded to a properly instructed AI, which did effectively the same steps I was doing. I also supervised each and every function and reviewed every single line of code AI produced. 

I really hope this will once and for good fix all of the complaints we're getting about the IDEs being unhappy when you `import valkey`. Or it can be at least a good starting point.

# Another important note

This change actually revealed quite a few bigger problems in the code base. The most common one is the fact that some of the async tests actually test the **sync** version of Valkey client. I will fix them slowly and gradually in the later PRs. This PR is large enough by itself.

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -2,7 +2,7 @@
 #, docs/examples, tests
 files = valkey
 check_untyped_defs = True
-follow_imports_for_stubs asyncio.= True
+follow_imports_for_stubs = True
 #disallow_any_decorated = True
 disallow_subclassing_any = True
 #disallow_untyped_calls = True
@@ -21,4 +21,24 @@ disallow_any_unimported = True
 
 [mypy-valkey.asyncio.lock]
 # TODO: Remove once locks has been rewritten
+ignore_errors = True
+
+[mypy-valkey.commands.bf.*]
+# Unsupported command family for now.
+ignore_errors = True
+
+[mypy-valkey.commands.search.*]
+# Unsupported command family for now.
+ignore_errors = True
+
+[mypy-valkey.commands.json.*]
+# Unsupported command family for now.
+ignore_errors = True
+
+[mypy-valkey.commands.graph.*]
+# Unsupported command family for now.
+ignore_errors = True
+
+[mypy-valkey.commands.timeseries.*]
+# Unsupported command family for now.
 ignore_errors = True
--- a/valkey/commands/core.py
+++ b/valkey/commands/core.py
@@ -13,6 +13,7 @@ from typing import (
     List,
     Literal,
     Mapping,
+    NoReturn,
     Optional,
     Sequence,
     Set,
@@ -29,7 +30,7 @@ from valkey.typing import (
     AnyStreamIdT,
     BitfieldOffsetT,
     ChannelT,
-    CommandsProtocol,
+    CommandsMixin,
     ConsumerT,
     EncodableT,
     ExpiryT,
@@ -52,7 +53,7 @@ if TYPE_CHECKING:
     from valkey.client import Valkey
 
 
-class ACLCommands(CommandsProtocol):
+class ACLCommands(CommandsMixin):
     """
     Valkey Access Control List (ACL) commands.
     see: https://valkey.io/topics/acl
@@ -389,7 +390,7 @@ class ACLCommands(CommandsProtocol):
 AsyncACLCommands = ACLCommands
 
 
-class ManagementCommands(CommandsProtocol):
+class ManagementCommands(CommandsMixin):
     """
     Valkey management commands
     """
@@ -1536,7 +1537,7 @@ class BitFieldOperation:
         return self.client.execute_command(*command)
 
 
-class BasicKeyCommands(CommandsProtocol):
+class BasicKeyCommands(CommandsMixin):
     """
     Valkey basic key-based commands
     """
@@ -2551,7 +2552,7 @@ class AsyncBasicKeyCommands(BasicKeyCommands):
         return super().unwatch()
 
 
-class ListCommands(CommandsProtocol):
+class ListCommands(CommandsMixin):
     """
     Valkey commands for List data type.
     see: https://valkey.io/topics/data-types#lists
@@ -2971,7 +2972,7 @@ class ListCommands(CommandsProtocol):
 AsyncListCommands = ListCommands
 
 
-class ScanCommands(CommandsProtocol):
+class ScanCommands(CommandsMixin):
     """
     Valkey SCAN commands.
     see: https://valkey.io/commands/scan
@@ -3307,7 +3308,7 @@ class AsyncScanCommands(ScanCommands):
                 yield d
 
 
-class SetCommands(CommandsProtocol):
+class SetCommands(CommandsMixin):
     """
     Valkey commands for Set data type.
     see: https://valkey.io/topics/data-types#sets
@@ -3406,7 +3407,9 @@ class SetCommands(CommandsProtocol):
         """
         return self.execute_command("SMEMBERS", name, keys=[name])
 
-    def smismember(self, name: str, values: List, *args: List) -> Union[
+    def smismember(
+        self, name: str, values: List, *args: List
+    ) -> Union[
         Awaitable[List[Union[Literal[0], Literal[1]]]],
         List[Union[Literal[0], Literal[1]]],
     ]:
@@ -3485,7 +3488,7 @@ class SetCommands(CommandsProtocol):
 AsyncSetCommands = SetCommands
 
 
-class StreamCommands(CommandsProtocol):
+class StreamCommands(CommandsMixin):
     """
     Valkey commands for Stream data type.
     see: https://valkey.io/topics/streams-intro
@@ -4094,7 +4097,7 @@ class StreamCommands(CommandsProtocol):
 AsyncStreamCommands = StreamCommands
 
 
-class SortedSetCommands(CommandsProtocol):
+class SortedSetCommands(CommandsMixin):
     """
     Valkey commands for Sorted Sets data type.
     see: https://valkey.io/topics/data-types-intro#valkey-sorted-sets
@@ -4893,7 +4896,7 @@ class SortedSetCommands(CommandsProtocol):
 AsyncSortedSetCommands = SortedSetCommands
 
 
-class HyperlogCommands(CommandsProtocol):
+class HyperlogCommands(CommandsMixin):
     """
     Valkey commands of HyperLogLogs data type.
     see: https://valkey.io/topics/data-types-intro#hyperloglogs
@@ -4928,7 +4931,7 @@ class HyperlogCommands(CommandsProtocol):
 AsyncHyperlogCommands = HyperlogCommands
 
 
-class HashCommands(CommandsProtocol):
+class HashCommands(CommandsMixin):
     """
     Valkey commands for Hash data type.
     see: https://valkey.io/topics/data-types-intro#valkey-hashes
@@ -5282,7 +5285,7 @@ class AsyncScript:
             return await client.evalsha(self.sha, len(keys), *args)
 
 
-class PubSubCommands(CommandsProtocol):
+class PubSubCommands(CommandsMixin):
     """
     Valkey PubSub commands.
     see https://valkey.io/topics/pubsub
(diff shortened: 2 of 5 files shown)


mkmkme:
@amirreza-sf80 would you be able to check this version in your valkey-py usage? Would the changes from this PR actually make your IDE happier? I would really appreciate if you can test this. 

Also pinging all of the people that reported a problem about typing: @kamikaze, @Rafiot,

amirreza-sf80:
hi 👋 
I'll check it out
it might take me some time to check everything tho as i have a few other things at hand


thanks for your effort here, type checking could help a lot of people 🙌

kamikaze:
will check, thanks

Rafiot:
This you so much for the work, I'll give it a shot.

mkmkme:
Hey fellas,

Thanks a lot for your responsiveness! Did any of you have a chance to try this out yet?

amirreza-sf80:
hi sorry for the delay 
i did a very basic check and both mypy and ty seemed happy
don't currently have anything bigger to check this against

Rafiot:
I played a bit with the branch from a fairly small project and it's great!

The only new type ignore I have to add is on calls to `Pipeline.execute()` which is really not blocking anything.

kurtmckee:
I'm concerned about the approach taken in this PR.

First, this significantly expands the number of errors that mypy throws (from 955 in `main` to 1,279 in this branch).

Second, the `.pyi` file introduced here is thousands of lines long -- it's tough to review as a single PR.

Third, adding a `.pyi` file appears to create a third touchpoint for code changes, in addition to having to maintain the sync and async code at the same time.

Finally, this doesn't do the work of removing the existing type annotations.

For these reasons, I harbor concerns that the approach taken in this PR may increase the maintenance burden required for project upkeep.

mkmkme:
> I'm concerned about the approach taken in this PR.
> 
> First, this significantly expands the number of errors that mypy throws (from 955 in `main` to 1,279 in this branch).
> 
> Second, the `.pyi` file introduced here is thousands of lines long -- it's tough to review as a single PR.
> 
> Third, adding a `.pyi` file appears to create a third touchpoint for code changes, in addition to having to maintain the sync and async code at the same time.
> 
> Finally, this doesn't do the work of removing the existing type annotations.
> 
> For these reasons, I harbor concerns that the approach taken in this PR may increase the maintenance burden required for project upkeep.

Considering the architecture of the project that was created long before the typing system in Python, this appears to be the cleanest approach. You can have a look at https://github.com/valkey-io/valkey-py/issues/164 and the description of the PR to see all the difficulties appearing related to the type system.

Adding `.pyi` might be a burden, but for now it's much more important to keep the project comfortable for its users rather than its developers.

As for the removing of the existing type annotations, this should not change anything for the users since AFAIK the type annotations for the API will be taken from `.pyi` first if existing. We will address the dev experience later. This is by far the larger low-hanging fruit for now.

kurtmckee:
> Adding `.pyi` might be a burden, but for now it's much more important to keep the project comfortable for its users rather than its developers.

My gut is that prioritizing the quality of the codebase and project infrastructure is a much more important concern. This PR appears to introduce untested type annotations on top of existing untested type annotations, and I'm concerned that it will take an even more significant effort to unwind the mypy errors in the codebase if this merges.

> the type annotations for the API will be taken from `.pyi` first

This further concerns me, because `.pyi` files are helpful in cases that don't apply to this repo [[PEP 484 notes on stub files](https://peps.python.org/pep-0484/#stub-files)]:

* Adding type annotations while allowing code to continue running in Python 2 and 3
* Adding type annotations to packages whose authors haven't added type annotations yet
* Packages where type annotations are used at runtime (this happens, for example, in Pydantic, whose runtime behavior swaps out `Field` definitions for types like `int` and `UUID`, and mypy chokes on the usage sites)

The stub file introduced in this PR doesn't appear to be necessary because anything that the `.pyi` file could solve can be written directly into the project. It adds a third location for untested type annotations. And, it increases the number of mypy errors by 33%.

I won't push further on this, but my suggestion is to close this PR and take a different approach. This approach doesn't look like it's introducing a solution to the type annotation problems.

mkmkme:
> > Adding `.pyi` might be a burden, but for now it's much more important to keep the project comfortable for its users rather than its developers.
> 
> My gut is that prioritizing the quality of the codebase and project infrastructure is a much more important concern. This PR appears to introduce untested type annotations on top of existing untested type annotations, and I'm concerned that it will take an even more significant effort to unwind the mypy errors in the codebase if this merges.
> 
> > the type annotations for the API will be taken from `.pyi` first
> 
> This further concerns me, because `.pyi` files are helpful in cases that don't apply to this repo [[PEP 484 notes on stub files](https://peps.python.org/pep-0484/#stub-files)]:
> 
> * Adding type annotations while allowing code to continue running in Python 2 and 3
> * Adding type annotations to packages whose authors haven't added type annotations yet
> * Packages where type annotations are used at runtime (this happens, for example, in Pydantic, whose runtime behavior swaps out `Field` definitions for types like `int` and `UUID`, and mypy chokes on the usage sites)
> 
> The stub file introduced in this PR doesn't appear to be necessary because anything that the `.pyi` file could solve can be written directly into the project. It adds a third location for untested type annotations. And, it increases the number of mypy errors by 33%.
> 
> I won't push further on this, but my suggestion is to close this PR and take a different approach. This approach doesn't look like it's introducing a solution to the type annotation problems.

Please re-read the description of the PR carefully. The current biggest problem related to the typing hints is that there are two different classes for sync and async valkey clients, and those are _implemented_ exactly the same, to the point where some classes are implemented like `AsyncFooCommands = FooCommands`. The only real difference is in function `execute_command` that is either implemented in a sync or async way.

Effectively that creates a linter error for each library call where an async function call would fail with "`int | Awaitable[int]` can't be `await`-ed" or "can't subscribe to`dict | Awaitable[dict]`".

You can see those errors if you open any test in the repo. And this PR solves exactly that.

If you have some better ideas of how to fix it, I would love to hear them. But AFAIK for now this is the only option we have.

kurtmckee:
I think that I have some understanding of the problem, but I don't yet understand how this solves the problem.

In particular, you've mentioned that there are linter errors (I think you're referring to mypy and IDEs, and I see those issues too, but please correct me if you're using a different linting tool).

However, this PR increases the number of mypy errors by 33% and doesn't introduce mypy to CI. Meanwhile, IDE linting cannot be tested. For those reasons, I'm not able to understand the disconnect between an increase in mypy errors and claims like "this PR solves the problem".

As for a different path to resolution, I submitted PR #282 (admittedly, without realizing this PR was already open). In #282, mypy is added to CI and is configured such that mypy ignores a number of problematic modules but throws no errors. This establishes a baseline and draws a line in the sand. It also allows for granular expansion of mypy's coverage of the package by removing modules from mypy's exclusions list and fixing the associated issues. I therefore anticipate that this will allow for incremental -- and tested -- improvements.

I think that we share the same goal, but I think this PR cannot be effectively reviewed due to its size, compounds the existing problem, and doesn't prevent the problem from continuing to increase (for lack of CI).

kurtmckee:
Perhaps it'll be helpful to ask: if this merges as-is, what would the workflow look like to fix type annotations going forward? Do you anticipate that it could be done module-by-module, or in relatively large chunks of work? Or, do you anticipate it would require a single, large PR again?

Are you anticipating that the `.pyi` file is a stop-gap until the types in the module can be completely solved, or does the `.pyi` file become the authoritative source for type annotations?

Finally, what would be required at each improvement step? For example, would the sync, async, and `.pyi` code all need to be touched at the same time, or would work happen in the sync and async code paths only (assuming that mypy would treat the `.pyi` file as authoritative and ignore everything happening in the runtime code)?

Graeme22:
Perhaps this approach that recently got added to redis-py could be considered? https://github.com/redis/redis-py/pull/4005

mkmkme:
> Perhaps this approach that recently got added to redis-py could be considered? [redis/redis-py#4005](https://github.com/redis/redis-py/pull/4005)

Thanks @Graeme22, this does look promising, and I'll have a look at this ASAP.

mkmkme:
While their approach looks promising, it still suffers from some older patterns like here: https://github.com/redis/redis-py/blob/master/redis/commands/core.py#L605

@kurtmckee you do make a good point. Perhaps the best approach can be to temporarily even create a separate package like `typesched-redis` used to be. I'll have a further look into the PR that was shared with me and will consider multiple different directions.

This issue has been the most painful points for us for way too long.

Graeme22:
> While their approach looks promising, it still suffers from some older patterns like here: https://github.com/redis/redis-py/blob/master/redis/commands/core.py#L605

I think you may have misread, look above that line and you should see the overloads

Graeme22:
> This issue has been the most painful points for us for way too long.

Tbh redis-py merging with aioredis in the first place was a massive mistake

mkmkme:
> > While their approach looks promising, it still suffers from some older patterns like here: https://github.com/redis/redis-py/blob/master/redis/commands/core.py#L605
> 
> I think you may have misread, look above that line and you should see the overloads

Ah yeah, you're right! Well this isn't too great, but it seems to work, so it can work for us, too. Thanks, I'm a bit too tired and read it carelessly...

>Tbh redis-py merging with aioredis in the first place was a massive mistake

It might be. But after all, it's not the only module in the world to suffer from the same "sync+async client" pattern and to suffer from similar consequences. I've even seen some modules trying to create a perfect framework to "write once, apply for sync & async", but haven't considered any of them for us yet.

PR #266, opened 2026-02-20
Add in min_connections to available arguments

Masses:
### Pull Request check-list

### Description of change

This change allows for setting a minimum number of connections that the client should initialize and keep-alive, similar to functionality other language libraries provide:https://github.com/valkey-io/valkey-java. This mostly helps for systems that are exposed to sudden changes in load, such as when behind a load balancer without a gradual warm up.

I have tested this in our dev environment by running load tests with and without this change. Without the change, p99 latency of valkey commands (for this load test, a FT.SEARCH) jumps to over 200 ms (as measured by the application) until load subsides. With the change, p99 latency is higher than a gradual warmup but remains under 20ms (as measured by the application). These load tests were measured on the same hardware running the application, and a single shard with 2 replicas (all replica CPU utilization was below 10% for entire load test)

First time contributing, so let me know if I did anything wrong! Also have a stronger background on the JVM, so if I'm missing python idioms let me know.

### DIFF
--- a/valkey/asyncio/client.py
+++ b/valkey/asyncio/client.py
@@ -226,6 +226,7 @@ class Valkey(
         ssl_min_version: Optional[ssl.TLSVersion] = None,
         ssl_ciphers: Optional[str] = None,
         max_connections: Optional[int] = None,
+        min_connections: int = 0,
         single_connection_client: bool = False,
         health_check_interval: int = 0,
         client_name: Optional[str] = None,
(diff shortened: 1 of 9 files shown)


valkey/asyncio/client.py
```diff
@@ -368,6 +377,8 @@ def __await__(self):
         return self.initialize().__await__()
 
     async def initialize(self: _ValkeyT) -> _ValkeyT:
+        if hasattr(self.connection_pool, "initialize"):
```
mkmkme: As far as I understand, you're adding `initialize` to `connection_pool` with the same commit, so this `if` statement should always be true? Is there actually a need for this condition then?

PR #268, opened 2026-02-28
fix(async-cluster): prevent reinitialization after explicit aclose()

### LINKED ISSUE #223: async cluster, connection after aclose
hi
one thing i'm noticing is that with the sync cluster client, after `close` is called, you can no longer operate on server, so `valkey.get("a")` raises an exception

but with the async version, you can still work 
so `await valkey.set("a", "b")` sets the valkeu in server and `await valkey.get("a")` will return the result from server

latest valkey-py version (latest commit from github)
valkey 8.1.3

mkmkme:
Make async `ValkeyCluster` treat user-initiated `aclose()` as terminal by tracking closed state and raising a clear `ValkeyClusterException` on further use. Preserve inte`rnal retry/reinitialize behavior by introducing a private reset path used by command and pipeline retry flows, and add regression tests for post-close command/initialize calls.

fixes #223

### Pull Request check-list

### DIFF
--- a/tests/test_asyncio/conftest.py
+++ b/tests/test_asyncio/conftest.py
@@ -78,7 +78,7 @@ async def create_valkey(request):
                 await client.aclose()
                 await client.connection_pool.disconnect()
             else:
-                if flushdb:
+                if flushdb and not getattr(client, "_closed", False):
                     try:
                         await client.flushdb(target_nodes="primaries")
                     except valkey.ConnectionError:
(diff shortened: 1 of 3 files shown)


mkmkme:
@amirreza-sf80 could you check if this fixes the issue for you please? Thanks!

mkmkme:
Huh that's interesting. Now some tests are failing to Valkey cluster being closed. Let's see

amirreza-sf80:
sorry the situation in my country is a bit hard
I'm barley connected to internet 
can't do much for now

bogdanp05 (approved):
lgtm

PR #273, opened 2026-03-04
Drop "-dev" from py314

kurtmckee:
### Pull Request check-list

### Description of change

This PR introduces the following changes:

* The "dev" suffix for Python 3.14 is dropped, as it's GA now

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -89,7 +89,15 @@ jobs:
        max-parallel: 15
        fail-fast: false
        matrix:
-         python-version: ['3.10', '3.11', '3.11.1', '3.12', '3.13', '3.14-dev', 'pypy-3.10', 'pypy-3.11']
+         python-version:
+          - '3.10'
+          - '3.11'
+          - '3.11.1'
+          - '3.12'
+          - '3.13'
+          - '3.14'
+          - 'pypy-3.10'
+          - 'pypy-3.11'
          test-type: ['standalone', 'cluster']
          connection-type: ['libvalkey', 'plain']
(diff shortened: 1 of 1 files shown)


mkmkme (changes requested):
Thanks a lot for your help! Could you please drop 3.15 for now? We will enable it closer to the release date (October 2026). For now the CI is annoyingly slow enough to add even more Python versions. Otherwise looks good!

kurtmckee:
Can do. By the way, is there a reason to have both `3.11` and `3.11.1` listed at the same time in the list of Python versions? If not, I can remove that as well.

mkmkme:
> Can do. By the way, is there a reason to have both `3.11` and `3.11.1` listed at the same time in the list of Python versions? If not, I can remove that as well.

Yes. There's a hack specific to pre-3.11.3 Python related to async: https://github.com/valkey-io/valkey-py/blob/cca695f6465afeb74787f85d4ab852d16f5b9750/tests/test_asyncio/test_pubsub.py#L8-L13

It's quite unfortunate to still have it and to still have to test it, but IIRC last time we checked some of the distros actually shipped Python 3.11 that predated 3.11.3, so we have to keep it.

I'll be more than happy to drop it once it's no longer needed :)

**EDIT:** I've noticed I pasted a link to a snippet in `tests`. In fact, there's an identical block in `valkey/asyncio/connection.py`.

kurtmckee:
@mkmkme I think I've addressed you feedback. Thanks for reviewing this PR!

mkmkme (approved):
Thanks, looks great!

PR #290, opened 2026-03-25
Fix resource warnings: unclosed sockets / SSL sockets / a PEM file

kurtmckee:
### Description of change

This fixes a number of resource warnings reported by pytest.

One exists in the test suite itself -- an unclosed PEM file -- but the rest are unclosed sockets and SSL sockets that are problems in the package itself.

The warnings shown below are representative:

```
ResourceWarning: unclosed <ssl.SSLSocket ...>
ResourceWarning: unclosed <socket.socket ...>
ResourceWarning: unclosed file <_io.BufferedReader name='dockers/stunnel/keys/server-cert.pem'>
```

> [!NOTE]
>
> This PR DOES NOT resolve all of the warnings that are thrown in the test suite!
>
> There is a lot of work to do on the [test suite's warnings](https://github.com/valkey-io/valkey-py/actions/runs/23520510469/job/68462725179#step:6:443), but this is an initial salvo that resolves the warnings that are displayed when running `make tests` locally.

### Pull Request check-list

### DIFF
--- a/tests/test_ssl.py
+++ b/tests/test_ssl.py
@@ -263,6 +263,8 @@ class TestSSL:
         ctx.use_certificate_file(self.CLIENT_CERT)
         ctx.use_privatekey_file(self.CLIENT_KEY)
 
+        with open(self.SERVER_CERT, "rb") as file:
+            server_cert = file.read()
         r = valkey.Valkey(
             host=p[0],
             port=p[1],
@@ -272,7 +274,7 @@ class TestSSL:
             ssl_cert_reqs="required",
             ssl_ca_certs=self.CA_CERT,
             ssl_ocsp_context=ctx,
-            ssl_ocsp_expected_cert=open(self.SERVER_CERT, "rb").read(),
+            ssl_ocsp_expected_cert=server_cert,
             ssl_validate_ocsp_stapled=True,
         )
 
--- a/valkey/connection.py
+++ b/valkey/connection.py
@@ -908,11 +908,20 @@ class SSLConnection(Connection):
             )
 
             #  need another socket
-            con = OpenSSL.SSL.Connection(staple_ctx, socket.socket())
-            con.request_ocsp()
-            con.connect((self.host, self.port))
-            con.do_handshake()
-            con.shutdown()
+            ocsp_socket = socket.socket()
+            try:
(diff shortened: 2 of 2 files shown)


bogdanp05 (approved):

kurtmckee:
The patch coverage warning occurs because the test suite doesn't test with the `[ocsp]` extra included, so the code changed here is never run by the test suite..._yet_. As noted in out-of-band communication, follow-up work is needed to make CI run the OCSP tests that will exercise these code paths.

kurtmckee:
_Rebased on `main` and force-pushed._

@bogdanp05 If this still looks good to you, would you be willing to merge?

bogdanp05:
@kurtmckee yep thanks for the ping, enabled auto-merge

kurtmckee:
I see that there's a hyper-specific CI failure (Python 3.12, RESP3, standalone mode with libvalkey).

I think there will be a scheduled CI run in a few hours against `main`; I check back later to see if `main` similarly fails or if this change has introduced this highly specific error.

mkmkme:
This was the reason I didn't merge it originally
image

bogdanp05:
@mkmkme I saw that one, but also noticed that it fires on a line that wasn't touched by this PR. Should we dismiss it and fix separately or patch it in this PR?

kurtmckee:
@mkmkme As I [noted above](https://github.com/valkey-io/valkey-py/pull/290#discussion_r3021471062), this is not new code -- it's simply indented in a `try` block. CodeQL is noticing that the lines were touched at all and is flagging them.

kurtmckee:
@bogdanp05 I don't think that the warnings should be addressed in this PR. Perhaps they should be addressed _first_, but I don't think it should happen in this PR, which is very targeted to reducing some of the existing test suite warnings.

mkmkme:
> @mkmkme I saw that one, but also noticed that it fires on a line that wasn't touched by this PR. Should we dismiss it and fix separately or patch it in this PR?

We could, but we also need to fix it eventually since this is an actual issue

bogdanp05:
I created https://github.com/valkey-io/valkey-py/issues/312 to keep track of it. Once you rebase, I can dismiss the error and merge.

kurtmckee:
_Rebased on `main` and force-pushed._

PR #302, opened 2026-04-22
Make skip_ifmodversion_lt skip not installed modules

bugant:
### Pull Request check-list

### Description of change

`skip_ifmodversion_lt` should not make a test to fail, if the needed module is not enabled.  Here the marker is updated to skip the test in that case, instead.


bogdanp05 (approved):
lgtm

PR #303, opened 2026-04-22
Fix test_acl_getuser_setuser for newer Valkey versions

bugant:
### Pull Request check-list

### Description of change

Newer Valkey versions return additional 'databases'/'alldbs' fields in ACL selector responses. Use `assert_resp_response_in` to accept both the old format (without `databases`) and the new format (with `databases: alldbs`).

### DIFF
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -331,11 +331,17 @@ class TestValkeyCommands:
         assert set(acl["keys"]) == {"~cache:*", "~objects:*"}
         assert len(acl["passwords"]) == 2
         assert set(acl["channels"]) == {"&message:*"}
-        assert_resp_response(
+        assert_resp_response_in(
             r,
             acl["selectors"],
-            [["commands", "-@all +set", "keys", "%W~app*", "channels", ""]],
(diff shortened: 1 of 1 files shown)


tests/test_commands.py
```diff
@@ -331,11 +331,17 @@ def teardown():
...
         assert set(acl["channels"]) == {"&message:*"}
-        assert_resp_response(
+        assert_resp_response_in(
             r,
             acl["selectors"],
-            [["commands", "-@all +set", "keys", "%W~app*", "channels", ""]],
-            [{"commands": "-@all +set", "keys": "%W~app*", "channels": ""}],
+            [
+                [["commands", "-@all +set", "keys", "%W~app*", "channels", ""]],
+                [["commands", "-@all +set", "keys", "%W~app*", "channels", "", "databases", "alldbs"]],
```
bogdanp05, line 339: I'm not sure about this change. The original test passes on 9.0.3, the latest stable release, but fails on 9.1.0-rc1. I don't think we should add test coverage for unstable releases, as they're subject to change.

PR #314, opened 2026-06-24
Fix CLUSTER MYSHARDID parsing with decode_responses=True

### LINKED ISSUE #309: Parse cluster myshardid decodes already decoded output
https://github.com/valkey-io/valkey-py/blob/470a39975cf7efd58792ec1eb2fbcbf55a01c1be/valkey/cluster.py#L122-L126
Could this check if output isn't already decoded?
```
>>> rc = Valkey(host='localhost', port=6379, decode_responses=True)
>>> rc.cluster_myshardid()
Traceback (most recent call last):
  File "<python-input-29>", line 1, in <module>
    rc.cluster_myshardid()
    ~~~~~~~~~~~~~~~~~~~~^^
  File "/home/rkozlo/Projekty/ansible_collections/rkozlo/valkey/.venv/lib64/python3.13/site-packages/valkey/commands/cluster.py", line 647, in cluster_myshardid
    return self.execute_command("CLUSTER MYSHARDID", target_nodes=target_nodes)
           ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rkozlo/Projekty/ansible_collections/rkozlo/valkey/.venv/lib64/python3.13/site-packages/valkey/cluster.py", line 1119, in execute_command
    raise e
  File "/home/rkozlo/Projekty/ansible_collections/rkozlo/valkey/.venv/lib64/python3.13/site-packages/valkey/cluster.py", line 1105, in execute_command
    res[node.name] = self._execute_command(node, *args, **kwargs)
                     ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rkozlo/Projekty/ansible_collections/rkozlo/valkey/.venv/lib64/python3.13/site-packages/valkey/cluster.py", line 1220, in _execute_command
    raise e
  File "/home/rkozlo/Projekty/ansible_collections/rkozlo/valkey/.venv/lib64/python3.13/site-packages/valkey/cluster.py", line 1161, in _execute_command
    response = self.cluster_response_callbacks[command](
        response, **kwargs
    )
  File "/home/rkozlo/Projekty/ansible_collections/rkozlo/valkey/.venv/lib64/python3.13/site-packages/valkey/cluster.py", line 126, in parse_cluster_myshardid
    return resp.decode("utf-8")
           ^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'decode'. Did you mean: 'encode'?

```

vipi-n:
## Summary
`parse_cluster_myshardid` unconditionally calls `resp.decode("utf-8")`.
When the client is created with `decode_responses=True`, the response is
already a `str`, so `cluster_myshardid()` raises
`AttributeError: 'str' object has no attribute 'decode'`.

Fixes #309

## Changes
- `valkey/cluster.py`: use the existing `str_if_bytes` helper (already
  imported in this module) instead of calling `.decode()` directly, so the
  parser works for both `decode_responses=False` (bytes) and
  `decode_responses=True` (str).
- `tests/test_cluster.py`: add a regression test covering both the bytes
  and already-decoded `str` inputs.

### DIFF
--- a/tests/test_cluster.py
+++ b/tests/test_cluster.py
@@ -1183,6 +1183,12 @@ class TestClusterValkeyCommands:
         myshardid = r.cluster_myshardid()
         assert isinstance(myshardid, str)
         assert len(myshardid) > 0
+        
+    def test_parse_cluster_myshardid_handles_decoded_input(self):
+        from valkey.cluster import parse_cluster_myshardid
+        assert parse_cluster_myshardid(b"abc123") == "abc123"
+        # decode_responses=True -> input is already str, must not raise
+        assert parse_cluster_myshardid("abc123") == "abc123"    
 
     def test_cluster_addslots(self, r):
         node = r.get_random_node()
--- a/tests/test_cluster.py
+++ b/tests/test_cluster.py
@@ -1179,17 +1179,12 @@ class TestClusterValkeyCommands:
                     assert attribute in attributes
 
     @skip_if_server_version_lt("7.2.0")
-    def test_cluster_myshardid(self, r):
-        myshardid = r.cluster_myshardid()
+    @pytest.mark.parametrize("client_fixture_name", ("r", "decoded_r"))
+    def test_cluster_myshardid(self, client_fixture_name, request):
+        client: ValkeyCluster = request.getfixturevalue(client_fixture_name)
+        myshardid = client.cluster_myshardid()
         assert isinstance(myshardid, str)
         assert len(myshardid) > 0
-
-    def test_parse_cluster_myshardid_handles_decoded_input(self):
-        from valkey.cluster import parse_cluster_myshardid
-
-        assert parse_cluster_myshardid(b"abc123") == "abc123"
-        # decode_responses=True -> input is already str, must not raise
-        assert parse_cluster_myshardid("abc123") == "abc123"
 
     def test_cluster_addslots(self, r):
         node = r.get_random_node()
--- a/tests/test_cluster.py
+++ b/tests/test_cluster.py
@@ -1183,9 +1183,10 @@ class TestClusterValkeyCommands:
         myshardid = r.cluster_myshardid()
         assert isinstance(myshardid, str)
         assert len(myshardid) > 0
-        
+
     def test_parse_cluster_myshardid_handles_decoded_input(self):
         from valkey.cluster import parse_cluster_myshardid
+
         assert parse_cluster_myshardid(b"abc123") == "abc123"
         # decode_responses=True -> input is already str, must not raise
         assert parse_cluster_myshardid("abc123") == "abc123"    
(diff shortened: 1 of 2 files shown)


tests/test_cluster.py
```diff
@@ -1184,6 +1184,13 @@ def test_cluster_myshardid(self, r):
         assert isinstance(myshardid, str)
         assert len(myshardid) > 0
 
+    def test_parse_cluster_myshardid_handles_decoded_input(self):
+        from valkey.cluster import parse_cluster_myshardid
+
+        assert parse_cluster_myshardid(b"abc123") == "abc123"
+        # decode_responses=True -> input is already str, must not raise
+        assert parse_cluster_myshardid("abc123") == "abc123"
+
```
kurtmckee: Please remove this test and modify the test directly above. This will ensure that the entire path is tested.

The test above currently uses a fixture named `r`, but this should be parametrized to use both `r` and `decoded_r`:


```python
    @skip_if_server_version_lt("7.2.0")
    @pytest.mark.parametrize("client_fixture_name", ("r", "decoded_r"))
    def test_cluster_myshardid(self, client_fixture_name, request):
        client: ValkeyCluster = request.getfixturevalue(client_fixture_name)
        myshardid = client.cluster_myshardid()
        assert isinstance(myshardid, str)
        assert len(myshardid) > 0
```

I ran this locally and confirmed that this demonstrated the failure before applying the patch, and passed after applying the patch.

valkey/cluster.py
```diff
@@ -123,7 +123,7 @@ def parse_cluster_myshardid(resp, **options):
     """
     Parse CLUSTER MYSHARDID response.
     """
-    return resp.decode("utf-8")
+    return str_if_bytes(resp)
```
kurtmckee: Please add type annotations here, since it's now known that this may be a string or bytes. Something like:

```python
def parse_cluster_myshardid(resp: str | bytes, **options: Any) -> str:
    """
    Parse CLUSTER MYSHARDID response.
    """
    return str_if_bytes(resp)
```

I tested this locally and confirmed that the test suite still passes, and that mypy continues to pass (though, admittedly, mypy isn't yet checking this module...but this will help prepare for that day).

kurtmckee (changes requested):
Hello @vipi-n! Thanks for addressing this issue!

I'm not an official developer on the project but have been working to improve the project's test suite infrastructure and wanted to provide feedback.

tests/test_cluster.py
vipi-n: Removed the separate test and parametrized test_cluster_myshardid over r and decoded_r as suggested, so the full path is covered.

valkey/cluster.py
vipi-n: Added resp: str | bytes, **options: Any and -> str annotations to parse_cluster_myshardid.

kurtmckee (approved):

mkmkme (commented):
The code itself looks good to me, thanks!

I am yet to see if we actually need to still return `str` or `bytes` in case `decode_responses=False`, but if we decide to return `str` unconditionally, I'll merge this.

This PR is good in a way it doesn't change the behaviour of the library while fixing the issue. I'm just unsure how we want to resolve the issue itself :)

PR #316, opened 2026-06-25
Remove testing using uvloop

kurtmckee:
### Pull Request check-list

### Description of change

This commit removes uvloop completely from valkey-py.

uvloop is being used exclusively in the test suite and in some benchmarking code.
However, its use incurs some costs:

* The test suite is being run twice in CI -- once without uvloop, and again with uvloop.
* When uvloop is used in the test suite, it throws deprecation warnings.

  Some warnings come from uvloop's own Python code when running on Python 3.14 and are due to upcoming Python 3.16 removals. Other warnings are from valkey-py's deprecated usage of uvloop.

To reduce test suite execution times,
and to reduce the burden of maintaining uvloop usage, uvloop is completely removed from the project.

When comparing [a CI run with this change](https://github.com/kurtmckee/pr-valkey-py/actions/runs/28170440039/usage) against [a recent scheduled CI run](https://github.com/valkey-io/valkey-py/actions/runs/28069029349/usage), the change is apparent:

| CI run | Wall clock time ("Duration") | Total run time |
|--------|--------|--------|
| `main` | `35m  4s` |  `5h 38m 28s` |
| This change | `30m 40s` | `4h 15m  8s` |

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -153,9 +153,6 @@ jobs:
            fi
            make devenv
            make ${{matrix.test-type}}-tests PROTOCOL=${{ matrix.protocol-version }}
-           if [[ "${{matrix.python-version}}" != pypy-* ]]; then
-            make ${{matrix.test-type}}-tests USE_UVLOOP=1 PROTOCOL=${{ matrix.protocol-version }}
-           fi
 
        - uses: actions/upload-artifact@v7
          if: success() || failure()
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,6 @@
-USE_UVLOOP             ?= 0
 PROTOCOL               ?= 2
 CLUSTER_URL            ?= valkey://localhost:16379/0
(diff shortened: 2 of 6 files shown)


mkmkme:
This is definitely not the way we want to go. `valkey-py` is being tested with `uvloop` because it can be run with `uvloop` event loop, and we want to make sure `valkey-py` works with it properly. That's why it is used in the tests.

`uvloop` itself is not a burden and never been something we wanted to remove.

The duration of CI is not a good justification for this. In my day job, I work with a project that can take more than 8 hours for a CI run, but so what? The number one priority for CI is to be thorough, not fast.

>Some warnings come from uvloop's own Python code when running on Python 3.14 and are due to upcoming Python 3.16 removals.

This is up to `uvloop` developers and has nothing to do with `valkey-py`

>Other warnings are from valkey-py's deprecated usage of uvloop.

You can create a ticket with these warnings listed, and we will fix this.

With that said, I'm closing this one.

PR #317, opened 2026-06-25
Migrate linting to use pre-commit hooks

kurtmckee:
### Pull Request check-list

### Description of change

This PR introduces the following changes:

* Introduce a pre-commit config.
* Migrate from running linters to running **fixers**.

For developers that use pre-commit, this change ensures that issues are caught **and fixed** immediately and locally, rather than running linters that merely notify developers what they should fix manually.

Note that there are a number of issues in the codebase to fix, and that there are many additional pre-commit hooks that will benefit the project. However, those changes can be introduced later. The pre-commit hooks and tools are configured to ensure that no issues have to be found and fixed until after this PR merges.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,71 @@
+# To execute the fixers in this file, run this command:
+#
+#   prek run -a
+#
+# To update the hook versions in this file, run this command:
+#
+#   prek auto-update --freeze --cooldown-days 20
+#
+
+default_language_version:
+  python: "python3.14"
+
+repos:
+  - repo: "meta"
+    hooks:
+      - id: "check-hooks-apply"
+      - id: "check-useless-excludes"
+
+  - repo: "https://github.com/pre-commit/pre-commit-hooks"
+    rev: "3e8a8703264a2f4a69428a0aa4dcb512790b2c8c"  # frozen: v6.0.0
+    hooks:
+      - id: "check-added-large-files"
+      - id: "check-merge-conflict"
+      - id: "check-yaml"
+# TODO: ENABLE end-of-file-fixer LATER
+#      - id: "end-of-file-fixer"
+      - id: "mixed-line-ending"
+        args:
+          - "--fix=lf"
+# TODO: ENABLE trailing-whitespace LATER
+#      - id: "trailing-whitespace"
+
+# TODO: ENABLE pyupgrade LATER
+#  - repo: "https://github.com/asottile/pyupgrade"
+#    rev: "75992aaa40730136014f34227e0135f63fc951b4"  # frozen: v3.21.2
+#    hooks:
+#      - id: "pyupgrade"
+#        name: "Enforce Python 3.10+ idioms"
+#        args: ["--py310-plus"]
+
+  - repo: "https://github.com/ikamensh/flynt/"
+    rev: "97be693bf18bc2f050667dd282d243e2824b81e2"  # frozen: 1.0.6
+    hooks:
+    -   id: "flynt"
+
+  - repo: "https://github.com/psf/black-pre-commit-mirror"
+    rev: "4160603246a6b365d4a2af661c6d71b0a0f50478"  # frozen: 26.5.1
+    hooks:
+      - id: "black-jupyter"
+
+  - repo: "https://github.com/pycqa/isort"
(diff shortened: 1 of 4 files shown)


mkmkme:
I'll need @bogdanp05's opinion here, but personally I dislike pre-commit hooks and don't want to switch to them. Running linters before committing is something every developer should do, it's just a matter of discipline. To make this easier for developers, there's even a `Makefile` that will do everything for you.

If there's code that violates style guides, CI will find it and flag it.

And even if Bogdan thinks we can start using pre-commit hooks for linting, I am **strongly** against pre-commit hooks that would *change the code automatically*. This is definitely a no-go for me. I should always be in control of what my commit changes and should be able to observe each change before I'm typing `git commit` byte-by-byte. Pre-commit hooks that fix the code for you violate that.

kurtmckee:
> I should always be in control of what my commit changes and should be able to observe each change before I'm typing `git commit` byte-by-byte. Pre-commit hooks that fix the code for you violate that.

This concern is unfounded.

First, developers must choose to install the pre-commit hooks as a git pre-commit hook using `pre-commit install`. That's an opt-in step that any developer can choose not to take.

Second, when a developer types `git commit` and the pre-commit hooks run, any changes made by fixers are never staged. The developer's staged changes are always -- ALWAYS -- untouched (and this fact, specifically, speaks to the concern you expressed above about byte-for-byte knowledge of what is about to be committed).

It's always incumbent on the developer to stage the changes that the tools have introduced, or fix the issues in some alternative way that is more palatable to the developer.

Third, if a developer has opted in to using the pre-commit hooks but strongly wants to commit changes that will fail the lint checks in CI, they can opt out by uninstalling the pre-commit hooks using `pre-commit uninstall`.

> there's even a `Makefile` that will do everything for you

The `Makefile` configures a number of tools to merely warn that linting has failed. This means that developers must re-find the problems by filename and line number and manually fix them. However, many of the tools have the ability to fix the problems automatically, which is very nice.

Thanks for reviewing the information above and reconsidering introducing these optional pre-commit hooks!

.pre-commit-config.yaml
```diff
@@ -0,0 +1,71 @@
...
+#
+#   prek run -a
+#
+# To update the hook versions in this file, run this command:
+#
+#   prek auto-update --freeze --cooldown-days 20
+#
+
+default_language_version:
+  python: "python3.14"
```
bogdanp05, line 11: this is 3.14 but in CI we use [3.13](https://github.com/valkey-io/valkey-py/blob/bdcbafa16320a260302a4279cc3fc4b3bcba174e/.github/workflows/integration.yaml#L48). They should ideally match

bogdanp05:
I'm not a fan of pre-commit hooks either because i find them too intrusive, but it looks like I won't have to change anything in my workflow after this PR. The only functional difference I see is that before `make linters` was read-only, now `make linters` also reformats. That's ok for me.

.pre-commit-config.yaml
kurtmckee: @bogdanp05 I've incorporated your feedback and rebased on `main`.

pyproject.toml
```diff
@@ -40,21 +40,17 @@ dependencies = [
...
 dev = [
-    'black[jupyter]',
     'cachetools',
-    'flake8-isort',
-    'flake8',
-    'flynt',
     'hatchling',
     'mock',
     'packaging>=20.4',
+    'prek',
```
amirreza-sf80: it is possible not to force a specific package for pre-commits
`.pre-commit-config.yaml` works with both `prek` and `pre-commit` tools so it is possible to let people decide what they want to use.

i would add `prek` in another group, other than `dev` so it's easy to install but not required

kurtmckee: `prek` was chosen here because something needed to exist to maintain Makefile-based linting, and `prek` is significantly faster than `pre-commit`. It's a superior choice _for use in the Makefile_.

You're welcome to install `pre-commit` and run `pre-commit install` locally for actual git pre-commit integration. However, I'm not going to add another dependency group just to provide options.

amirreza-sf80: adding dependency groups doesn't cause any overhead :)
i don't see why it would be considered a bad thing

i use prek myself, that's not the issue

if we add a package to `dev` people who want to contribute have to install it, but prek is not a required package to work in valkey-py, why should be required to install?

PR #319, opened 2026-06-27
Type annotate `valkey.utils`

kurtmckee:
### Pull Request check-list

### Description of change

This PR adds type annotations to `valkey.utils`. During this work, several functions were found to be unnecessary:

* The `pipeline()` context manager is not used anywhere in the codebase and is not documented anywhere.
* `list_keys_to_dict()` is an unnecessary pass-through function to `dict.fromkeys()`.
* `from_url()` is a pass-through function to `Valkey.from_url()`, but it creates a circular import.

Running `mypy --strict` shows that there are no remaining errors in `valkey.utils`.

### DIFF
--- a/valkey/utils.py
+++ b/valkey/utils.py
@@ -1,7 +1,6 @@
 import logging
-from contextlib import contextmanager
 from functools import wraps
-from typing import Any, Dict, Mapping, Union
+from typing import Any, Callable, Iterable, TypeVar, overload
 
 try:
     import libvalkey  # noqa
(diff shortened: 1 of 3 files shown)


valkey/utils.py
```diff
@@ -27,37 +26,28 @@
 
 from importlib import metadata
 
+T = TypeVar("T")
 
-def from_url(url, **kwargs):
```
bogdanp05, line 31: can we keep this function as well as `list_keys_to_dict` and `pipeline`? These are public function so we shouldn't break compatibility unless we have a good reason

valkey/utils.py
```diff
@@ -27,37 +26,28 @@
...
     )
 
 
-def safe_str(value):
+def safe_str(value: Any) -> str:
     return str(str_if_bytes(value))
 
 
-def dict_merge(*dicts: Mapping[str, Any]) -> Dict[str, Any]:
+def dict_merge(*dicts: dict[str, Any]) -> dict[str, Any]:
```
bogdanp05, line 50:
```suggestion
def dict_merge(*dicts: Mapping[str, Any]) -> dict[str, Any]:
```

valkey/utils.py
```diff
@@ -27,37 +26,28 @@
...
+def str_if_bytes(value: bytes) -> str: ...
 
 
-@contextmanager
-def pipeline(valkey_obj):
-    p = valkey_obj.pipeline()
-    yield p
-    p.execute()
+@overload
+def str_if_bytes(value: T) -> T: ...
```
bogdanp05, line 37: I think we should remove these overloads. 
If we pass `Any` to `str_if_bytes`, then mypy will evaluate the return as `Any`. But if we get rid of the overloads, and pass `Any` to `str_if_bytes`, then mypy will evaluate the return as `str | Any` which is preferable.

bogdanp05 (changes requested):

valkey/utils.py
kurtmckee: The overloads cannot be removed.

They reflect current usage of the function throughout the codebase, which in some cases may be e.g. lists (at least, according to the type annotations of the `.read_response()` methods' return values, which are passed directly to `str_if_bytes`.

bogdanp05: True, I missed that. Ok they should stay then

valkey/utils.py
kurtmckee: I've added these back. However, these utility functions are not used anywhere in the codebase and provide no discernible value to callers.

For example, `list_keys_to_dict()`:

```diff
- d = list_keys_to_dict(i, v)
+ d = dict.from_keys(i, v)
```

I advocate for their swift removal.

kurtmckee:
@bogdanp05 I've updated the PR based on your feedback. Thanks for reviewing this work!

bogdanp05 (dismissed):

kurtmckee:
Looks like vulture doesn't understand `if TYPE_CHECKING` blocks / string type annotations. I'll get this updated.

kurtmckee:
@bogdanp05 I've added the two lines to `whitelist.py` so vulture will ignore the imports used for type annotations.

bogdanp05 (approved):

PR #326, opened 2026-07-01
fix: route FT.SEARCH and search read commands to replicas when read_from_replicas=True

KarthikSubbarao:
### Pull Request check-list

### Description of change

fix: route FT.SEARCH and search read commands to replicas when read_from_replicas=True

Previously, all FT.* commands were lumped into a single SEARCH_COMMANDS list and unconditionally routed to the cluster's default_node (always a primary). This meant read_from_replicas=True had no effect on FT.SEARCH, FT.AGGREGATE, and other read-only search commands.

This change:
1. Splits SEARCH_COMMANDS into SEARCH_WRITE_COMMANDS (FT.CREATE, FT.ALTER, FT.DROPINDEX, etc.) and SEARCH_READ_COMMANDS (FT.SEARCH, FT.AGGREGATE, FT.INFO, etc.)
2. Routes write commands to default_node (primary) as before
3. Routes read commands to a random node (primary or replica) when read_from_replicas=True, enabling load distribution
4. Adds search read commands to READ_COMMANDS frozenset for proper classification throughout the codebase
5. Maintains backward-compatible SEARCH_COMMANDS as union of both lists

In valkey-search cluster mode with the coordinator (--use-coordinator), FT.SEARCH can execute on any node (primary or replica) because:
- The command is registered with the 'readonly' flag
- The coordinator gRPC server runs on all nodes
- The receiving node fans out to other shards regardless of its role
- Replicas maintain full search indexes via replication

Tested with valkey-search 9.0.1 in a 3-shard CME cluster confirming FT.SEARCH works on all replica nodes with cross-shard fanout.

### DIFF
--- a/tests/test_search_replica_routing.py
+++ b/tests/test_search_replica_routing.py
@@ -1,0 +1,81 @@
+"""
+Unit tests for FT.SEARCH replica routing fix.
+
+Verifies that search read commands are routed to replicas when
+read_from_replicas=True, while write commands always go to primaries.
+"""
+
+import unittest
+from unittest.mock import MagicMock, patch
+
+from valkey.cluster import AbstractValkeyCluster, ClusterNode, ValkeyCluster
+
+
+class TestSearchCommandClassification(unittest.TestCase):
+    """Test that search commands are correctly classified."""
+
+    def test_no_overlap_between_write_and_read(self):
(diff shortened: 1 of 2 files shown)


KarthikSubbarao:
@mkmkme - Hello! You can take a look when you get a chance. Thanks!

PR #332, opened 2026-08-13
Fix RuntimeError when the node dict is mutated during initialize()

EugeneChung:
### Pull Request check-list

### Description of change

`NodesManager.initialize()` iterates `self.startup_nodes.values()` directly. With `dynamic_startup_nodes` enabled (the default), the tail of `initialize()` runs

```python
if self._dynamic_startup_nodes:
    # Populate the startup nodes with all discovered nodes
    self.startup_nodes = tmp_nodes_cache
```

so from that point on `startup_nodes` and `nodes_cache` are the **same dict object**. Anything that mutates `nodes_cache` also mutates `startup_nodes`.

Two paths mutate that dict in place, and both are reached from ordinary command execution on other threads:

- `NodesManager._update_moved_slots()` — `self.nodes_cache[redirected_node.name] = redirected_node`, when a MOVED redirect names a node not yet in the cache.
- `ValkeyCluster._execute_command()`, on `ConnectionError`/`TimeoutError` — `self.nodes_manager.startup_nodes.pop(target_node.name, None)` followed by `nodes_manager.initialize()`.

So one thread can be inside `initialize()` walking `startup_nodes` while another inserts or pops, and the client raises

```
RuntimeError: dictionary changed size during iteration
```

out of whatever user command happened to trigger the re-initialization. `RuntimeError` is neither `ValkeyError` nor `ValkeyClusterException`, so callers that catch the library's exception hierarchy cannot absorb it.

#### How we hit it

A production ElastiCache Valkey cluster (18 shards, cluster mode enabled) was upgraded in place, which fails the primary of each shard over in batches. Every worker re-initialized at once while MOVED redirects were still flowing. 47 requests failed with the traceback below over a ~9 minute window; each one surfaced as a 500 rather than a handled cluster error.

```
valkey.exceptions.ClusterDownError: The cluster is down

During handling of the above exception, another exception occurred:
  ...
  valkey/cluster.py  initialize
RuntimeError: dictionary changed size during iteration
```

An engine upgrade is not special here — any primary failover, resharding, or slot migration produces the same combination of MOVED traffic and connection errors.

#### Fix

Iterate over a snapshot in `initialize()` and in the other methods that walk `nodes_cache` (`get_nodes_by_server_type`, `close`, `flush_cache`, `delete_command_from_cache`, `invalidate_key_from_cache`). `list(d.values())` is atomic with respect to other threads, so the loop no longer observes a concurrent resize.

One related line: inside the loop, `self.startup_nodes[startup_node.name].valkey_connection = r` becomes `startup_node.valkey_connection = r`. It is the same object, and it drops a lookup that would `KeyError` if another thread popped that name in the meantime.

This does not make `NodesManager` fully thread-safe — it removes this specific crash. Locking the topology state is a larger change.

#### Test

`tests/test_cluster.py::TestNodesManager::test_init_slots_cache_concurrent_node_mutation`

Deterministic and single-threaded: a mocked `CLUSTER SLOTS` handler inserts into `nodes_cache` and then raises `ConnectionError`, which is what a just-failed-over node looks like. The insert mutates the dict `initialize()` is iterating, and the raise makes the loop advance to the next startup node — where the unpatched code raises `RuntimeError`. (The loop `break`s early on full coverage, so a reachable first node never reaches a second iteration; the failure is what exposes it.)

Verified in both directions on this branch: without the `cluster.py` change the test reproduces `RuntimeError: dictionary changed size during iteration`; with it, the second startup node serves a full 16384-slot cache.

### DIFF
--- a/tests/test_cluster.py
+++ b/tests/test_cluster.py
@@ -2668,6 +2668,72 @@ class TestNodesManager:
                 "startup_nodes could not agree on a valid slots cache"
             ), str(ex.value)
 
+    def test_init_slots_cache_concurrent_node_mutation(self):
+        """
+        Test that initialize() tolerates the node dict being mutated while it
+        is iterating.
+
+        With dynamic_startup_nodes (the default) initialize() ends with
+        `self.startup_nodes = tmp_nodes_cache`, so startup_nodes and
+        nodes_cache become the same dict. A MOVED redirect handled on another
+        thread then inserts into the very dict this loop is walking, raising
+        "RuntimeError: dictionary changed size during iteration".
+
+        The loop only advances past its first startup node when that node is
+        unreachable, which is the state a failover leaves behind.
+        """
+        cluster_slots = [
+            [0, 8191, ["127.0.0.1", 7000]],
+            [8192, 16383, ["127.0.0.1", 7001]],
+        ]
+        failed_over = []
+
+        with patch.object(
+            NodesManager, "create_valkey_node", autospec=True
+        ) as create_valkey_node:
+
+            def create_mocked_valkey_node(nodes_manager, host, port, **kwargs):
+                r_node = Valkey(host=host, port=port)
+
+                def execute_command(*args, **kwargs):
+                    if args[0] == "CLUSTER SLOTS":
+                        if failed_over and port == 7000:
+                            # Stand in for _update_moved_slots() inserting a
+                            # redirected node from another thread, while this
+                            # node is the one that just failed over.
+                            node = ClusterNode("127.0.0.2", port)
+                            nodes_manager.nodes_cache[node.name] = node
+                            raise ConnectionError(f"{host}:{port} is down")
+                        return cluster_slots
+                    elif args[0] == "INFO":
+                        return {"cluster_enabled": True}
+                    elif args[1] == "cluster-require-full-coverage":
+                        return {"cluster-require-full-coverage": "yes"}
+                    else:
+                        return None
+
+                r_node.execute_command = execute_command
+                return r_node
+
+            create_valkey_node.side_effect = create_mocked_valkey_node
+
+            rc = ValkeyCluster(
+                startup_nodes=[
+                    ClusterNode("127.0.0.1", 7000),
+                    ClusterNode("127.0.0.1", 7001),
+                ]
+            )
+            n_manager = rc.nodes_manager
+            # dynamic_startup_nodes aliases the two dicts to one object
+            assert n_manager.startup_nodes is n_manager.nodes_cache
+
+            failed_over.append(True)
+            # Must not raise RuntimeError; the second startup node still
+            # serves a full slots cache.
+            n_manager.initialize()
+
+        assert len(n_manager.slots_cache) == VALKEY_CLUSTER_HASH_SLOTS
+
     def test_cluster_one_instance(self):
         """
         If the cluster exists of only 1 node then there is some hacks that must
--- a/valkey/cluster.py
+++ b/valkey/cluster.py
@@ -1464,7 +1464,7 @@ class NodesManager:
         """
         return [
             node
-            for node in self.nodes_cache.values()
+            for node in list(self.nodes_cache.values())
             if node.server_type == server_type
         ]
 
@@ -1536,7 +1536,10 @@ class NodesManager:
         fully_covered = False
         kwargs = self.connection_kwargs
         exception = None
-        for startup_node in self.startup_nodes.values():
+        # Iterate over a snapshot: another thread handling a MOVED redirect or a
+        # node failure may insert into or pop from this dict while we are here,
+        # and with dynamic_startup_nodes it is the same object as nodes_cache.
+        for startup_node in list(self.startup_nodes.values()):
             try:
                 if startup_node.valkey_connection:
                     r = startup_node.valkey_connection
@@ -1545,7 +1548,7 @@ class NodesManager:
                     r = self.create_valkey_node(
                         startup_node.host, startup_node.port, **kwargs
                     )
-                    self.startup_nodes[startup_node.name].valkey_connection = r
+                    startup_node.valkey_connection = r
                 # Make sure cluster mode is enabled on this node
                 try:
                     cluster_slots = str_if_bytes(r.execute_command("CLUSTER SLOTS"))
@@ -1658,7 +1661,7 @@ class NodesManager:
 
     def close(self):
         self.default_node = None
-        for node in self.nodes_cache.values():
+        for node in list(self.nodes_cache.values()):
             if node.valkey_connection:
                 node.valkey_connection.close()
 
@@ -1680,15 +1683,15 @@ class NodesManager:
         return host, port
 
     def flush_cache(self):
-        for node in self.nodes_cache.values():
+        for node in list(self.nodes_cache.values()):
             node.flush_cache()
 
     def delete_command_from_cache(self, command):
-        for node in self.nodes_cache.values():
+        for node in list(self.nodes_cache.values()):
             node.delete_command_from_cache(command)
 
     def invalidate_key_from_cache(self, key):
-        for node in self.nodes_cache.values():
+        for node in list(self.nodes_cache.values()):
             node.invalidate_key_from_cache(key)
 
 


EugeneChung:
## Repro against a real cluster

The test in this PR is deterministic and single-threaded by design. Here is the
complementary end-to-end version: the crash on this branch's merge-base, and its absence on
the branch head, against a real 6-shard Valkey cluster under concurrent load. No mocks.

Sources are inline at the bottom of this comment — three files, no dependencies beyond
`podman` (or `docker`) and Python ≥ 3.10.

```
./run.sh all            # up -> before -> fresh cluster -> after
./run.sh up             # 6 shards x 1 replica on 127.0.0.1:7300-7311
./run.sh before         # merge-base of this branch -> expect RuntimeError
./run.sh after          # branch head              -> expect clean
./run.sh down
```

`run.sh` exports the two source trees with `git archive` from the merge-base and the branch
head, so the two runs differ only by this PR's commits. The working tree is never touched.

### What it does

`cluster-up.sh` starts 12 `valkey-server` processes inside one `valkey/valkey:9.0`
container — 6 primaries, 6 replicas, `cluster-node-timeout 2000`, and
`cluster-announce-ip 127.0.0.1` so the published ports work from the host.

`repro.py` runs 48 client threads doing `SET`/`GET` on random keys through one
`ValkeyCluster`, while chaos threads `SIGSTOP` up to 2 current primaries at a time for 6s
and then `SIGCONT` them.

A frozen primary is what a just-failed-over node looks like from the client, and it feeds
both mutation paths at once:

- commands routed to it hang until `socket_timeout`, so `_execute_command()` runs
  `startup_nodes.pop(...)` and then `initialize()`
- the surviving nodes fail it over after `cluster-node-timeout`, so MOVED redirects flow and
  `_update_moved_slots()` inserts into `nodes_cache`

`dynamic_startup_nodes` is left at its default, so `startup_nodes` and `nodes_cache` are the
same object and both mutations land in the dict another thread is iterating.

`SIGSTOP` rather than killing the process is the part that matters. A killed node refuses
connections instantly, so `initialize()` skips it in well under a millisecond. A stopped
process still completes the TCP handshake out of the listen backlog, so `CLUSTER SLOTS`
blocks for the full `socket_timeout`. That widens the loop from sub-millisecond to seconds,
which is what makes the race land in tens of seconds instead of hours.

### Result

macOS 15 / podman 5.4.1 / valkey 9.0 / Python 3.13, defaults (48 threads, 6s freezes).

```
[before] valkey-py from .build/before/valkey/__init__.py
[before] connected, 6 primaries
[before] 24.2s elapsed
     20239  ok
       189  ClusterDownError
        38  RuntimeError: dictionary changed size during iteration
         8  freeze

[after] valkey-py from .build/after/valkey/__init__.py
[after] connected, 6 primaries
[after] 120.6s elapsed
     83139  ok
       186  ClusterDownError
        36  freeze
        30  TimeoutError

==============================================
before (merge-base): REPRODUCED RuntimeError
after  (branch head): no RuntimeError
==============================================
```

`before` reproduces in 20–40s and kills whichever worker thread it lands on. The traceback
matches the production one in the description:

```
Traceback (most recent call last):
  File "/Users/euigeun/project/opensources/valkey-py/repro-332/.build/before/valkey/cluster.py", line 1159, in _execute_command
    response = valkey_node.parse_response(connection, command, **kwargs)
  File "/Users/euigeun/project/opensources/valkey-py/repro-332/.build/before/valkey/client.py", line 605, in parse_response
    response = connection.read_response()
  File "/Users/euigeun/project/opensources/valkey-py/repro-332/.build/before/valkey/connection.py", line 588, in read_response
    raise response
valkey.exceptions.ClusterDownError: The cluster is down

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/euigeun/project/opensources/valkey-py/repro-332/repro.py", line 45, in worker
    rc.set(key, b"v")
    ~~~~~~^^^^^^^^^^^
  File "/Users/euigeun/project/opensources/valkey-py/repro-332/.build/before/valkey/commands/core.py", line 2326, in set
    return self.execute_command("SET", *pieces, **options)
           ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/euigeun/project/opensources/valkey-py/repro-332/.build/before/valkey/cluster.py", line 1119, in execute_command
    raise e
  File "/Users/euigeun/project/opensources/valkey-py/repro-332/.build/before/valkey/cluster.py", line 1105, in execute_command
    res[node.name] = self._execute_command(node, *args, **kwargs)
                     ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/euigeun/project/opensources/valkey-py/repro-332/.build/before/valkey/cluster.py", line 1213, in _execute_command
    self.nodes_manager.initialize()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/euigeun/project/opensources/valkey-py/repro-332/.build/before/valkey/cluster.py", line 1541, in initialize
    for startup_node in self.startup_nodes.values():
                        ~~~~~~~~~~~~~~~~~~~~~~~~~^^
RuntimeError: dictionary changed size during iteration
```

`_execute_command()` calls `initialize()` from three exception handlers. Two of them were
observed raising across runs: the `ClusterDownError` handler (the traceback above), and the
`ConnectionError`/`TimeoutError` handler that does the `startup_nodes.pop()`. The
`MovedError`-past-`reinitialize_steps` handler was not hit in these runs.

`after` ran 120s and 83k operations through the same churn — 36 freezes, the same
`ClusterDownError` and `TimeoutError` traffic — with no `RuntimeError`. Those remaining
errors are the expected, catchable ones; they are `ValkeyError` subclasses, which is what
this PR does not change.

Nothing here is upgrade-specific. Setting `VALKEY_IMAGE=valkey/valkey:9.1` between the two
runs turns it into a version upgrade, but the crash only needs failover churn — freezing
primaries on a single version is enough.

### Sources

<details>
<summary><code>run.sh</code></summary>

```bash
#!/usr/bin/env bash
# Repro harness for valkey-py PR #332.
#   ./run.sh up        - start a 6-shard valkey cluster on 127.0.0.1:7000-7011
#   ./run.sh before    - run the load+chaos against the PR's merge-base (expect RuntimeError)
#   ./run.sh after     - run it against the PR head (expect clean)
#   ./run.sh all       - up, before, fresh cluster, after
#   ./run.sh down      - tear down
set -euo pipefail

HERE=$(cd "$(dirname "$0")" && pwd)
REPO=$(cd "$HERE/.." && pwd)
NAME=${NAME:-valkey-repro-332}
IMAGE=${VALKEY_IMAGE:-valkey/valkey:9.0}
BUILD=$HERE/.build
PYTHON=${PYTHON:-python3}
PR_BRANCH=${PR_BRANCH:-fix/nodes-manager-concurrent-mutation}
# 7000 is taken by AirPlay Receiver on macOS.
START_PORT=${START_PORT:-7300}
NODES=${NODES:-12}
END_PORT=$((START_PORT + NODES - 1))

up() {
  podman rm -f "$NAME" >/dev/null 2>&1 || true
  podman run -d --name "$NAME" --user 0 \
    -p "$START_PORT-$END_PORT:$START_PORT-$END_PORT" \
    -e "START_PORT=$START_PORT" -e "NODES=$NODES" \
    --entrypoint /bin/sh "$IMAGE" -c "$(cat "$HERE/cluster-up.sh")" >/dev/null
  printf 'waiting for cluster on %s' "$IMAGE"
  for _ in $(seq 90); do
    if podman logs "$NAME" 2>&1 | grep -q 'CLUSTER READY'; then
      echo ' ok'
      return 0
    fi
    printf '.'
    sleep 1
  done
  echo ' FAILED'
  podman logs "$NAME" 2>&1 | tail -40
  exit 1
}

trees() {
  [ -d "$BUILD/after/valkey" ] && [ -d "$BUILD/before/valkey" ] && return 0
  rm -rf "$BUILD"
  mkdir -p "$BUILD/before" "$BUILD/after"
  git -C "$REPO" fetch -q origin "$PR_BRANCH"
  local head base
  head=$(git -C "$REPO" rev-parse FETCH_HEAD)
  base=$(git -C "$REPO" merge-base FETCH_HEAD origin/main)
  echo "before = $base (merge-base)"
  echo "after  = $head ($PR_BRANCH)"
  git -C "$REPO" archive "$base" | tar -x -C "$BUILD/before"
  git -C "$REPO" archive "$head" | tar -x -C "$BUILD/after"
}

variant() {
  local v=$1
  shift
  trees
  PYTHONPATH="$BUILD/$v" "$PYTHON" "$HERE/repro.py" \
    --container "$NAME" --label "$v" \
    --start-port "$START_PORT" --nodes "$NODES" "$@"
}

case "${1:-all}" in
  up) up ;;
  down) podman rm -f "$NAME" ;;
  trees) trees ;;
  before | after)
    v=$1
    shift
    variant "$v" "$@"
    ;;
  all)
    shift || true
    up
    rc_before=0
    variant before "$@" || rc_before=$?
    up # the cluster has been mauled; start clean for the control run
    rc_after=0
    variant after "$@" || rc_after=$?
    echo
    echo "=============================================="
    echo "before (merge-base): $([ $rc_before -ne 0 ] && echo 'REPRODUCED RuntimeError' || echo 'no RuntimeError')"
    echo "after  (PR head)   : $([ $rc_after -ne 0 ] && echo 'REPRODUCED RuntimeError' || echo 'no RuntimeError')"
    echo "=============================================="
    [ $rc_before -ne 0 ] && [ $rc_after -eq 0 ]
    ;;
  *)
    echo "usage: $0 {up|down|before|after|all} [repro.py args...]" >&2
    exit 2
    ;;
esac
```

</details>

<details>
<summary><code>cluster-up.sh</code></summary>

```sh
#!/bin/sh
# Runs INSIDE the container. 12 valkey nodes on 127.0.0.1:7000-7011 => 6 shards x 1 replica.
set -e

START=${START_PORT:-7300}
END=$((START + ${NODES:-12} - 1))

for p in $(seq $START $END); do
  mkdir -p /data/$p
  valkey-server \
    --port $p \
    --bind 0.0.0.0 \
    --protected-mode no \
    --cluster-enabled yes \
    --cluster-config-file nodes.conf \
    --cluster-node-timeout 2000 \
    --cluster-announce-ip 127.0.0.1 \
    --dir /data/$p \
    --save '' \
    --appendonly no \
    --logfile /data/$p/valkey.log &
  echo $! > /data/$p.pid
done

sleep 2
echo yes | valkey-cli --cluster create $(seq -f 127.0.0.1:%g $START $END) --cluster-replicas 1

# Wait until every node agrees the cluster is up.
for _ in $(seq 30); do
  if [ "$(valkey-cli -p $START cluster info | grep -c 'cluster_state:ok')" = "1" ]; then
    echo "CLUSTER READY"
    break
  fi
  sleep 1
done

wait
```

</details>

<details>
<summary><code>repro.py</code></summary>

```python
#!/usr/bin/env python3
"""Repro harness for valkey-py PR #332.

Drives a real 6-shard Valkey cluster with N client threads while a chaos thread
SIGSTOPs primaries. A frozen primary gives the client both halves of the race:

  * commands to it hang until socket_timeout -> ConnectionError/TimeoutError ->
    ValkeyCluster._execute_command() does startup_nodes.pop(...) then initialize()
  * the surviving nodes fail it over after cluster-node-timeout -> MOVED storm ->
    NodesManager._update_moved_slots() inserts into nodes_cache

With dynamic_startup_nodes (the default) startup_nodes IS nodes_cache, so those
mutations land in the dict that another thread is iterating inside initialize().

Unpatched -> RuntimeError: dictionary changed size during iteration.
"""

import argparse
import random
import subprocess
import sys
import threading
import time
import traceback
from collections import Counter

from valkey.cluster import ClusterNode, ValkeyCluster

STOP = threading.Event()
LOCK = threading.Lock()
COUNTS = Counter()
CRASH = []
FROZEN = set()


def bump(name, n=1):
    with LOCK:
        COUNTS[name] += n


def worker(rc):
    while not STOP.is_set():
        key = f"k:{random.randrange(1 << 20)}"
        try:
            rc.set(key, b"v")
            rc.get(key)
            bump("ok")
        except RuntimeError as e:
            # The bug. Not a ValkeyError, so callers catching the library
            # hierarchy cannot absorb it.
            bump(f"RuntimeError: {e}")
            with LOCK:
                if not CRASH:
                    CRASH.append(traceback.format_exc())
            STOP.set()
            return
        except Exception as e:
            bump(type(e).__name__)


def exec_in(container, cmd):
    subprocess.run(
        ["podman", "exec", container, "sh", "-c", cmd],
        check=False,
        capture_output=True,
    )


def pick_primary(rc, ports):
    try:
        primaries = [n.port for n in rc.get_primaries()]
    except Exception:
        primaries = ports
    with LOCK:
        free = [p for p in primaries if p not in FROZEN]
    return random.choice(free) if free else None


def chaos(rc, container, ports, freeze, gap, max_frozen):
    while not STOP.is_set():
        STOP.wait(random.uniform(0, gap))
        if STOP.is_set():
            return
        with LOCK:
            if len(FROZEN) >= max_frozen:
                busy = True
            else:
                busy = False
        if busy:
            continue

        victim = pick_primary(rc, ports)
        if victim is None:
            continue
        with LOCK:
            if victim in FROZEN or len(FROZEN) >= max_frozen:
                continue
            FROZEN.add(victim)
        try:
            exec_in(container, f"kill -STOP $(cat /data/{victim}.pid)")
            bump("freeze")
            STOP.wait(freeze)
        finally:
            exec_in(container, f"kill -CONT $(cat /data/{victim}.pid)")
            with LOCK:
                FROZEN.discard(victim)


def main():
    ap = argparse.ArgumentParser()
    ap.add_argument("--container", default="valkey-repro-332")
    ap.add_argument("--start-port", type=int, default=7300)
    ap.add_argument("--nodes", type=int, default=12)
    ap.add_argument("--threads", type=int, default=48)
    ap.add_argument("--duration", type=float, default=90)
    ap.add_argument("--freeze", type=float, default=6.0)
    ap.add_argument("--gap", type=float, default=2.0)
    ap.add_argument("--max-frozen", type=int, default=2)
    ap.add_argument("--socket-timeout", type=float, default=2.0)
    ap.add_argument("--label", default="")
    args = ap.parse_args()

    import valkey

    ports = list(range(args.start_port, args.start_port + args.nodes))
    print(f"[{args.label or 'run'}] valkey-py from {valkey.__file__}")

    rc = ValkeyCluster(
        startup_nodes=[ClusterNode("127.0.0.1", p) for p in ports],
        socket_timeout=args.socket_timeout,
        socket_connect_timeout=args.socket_timeout,
        # dynamic_startup_nodes=True is the default and is what makes
        # startup_nodes and nodes_cache the same object.
    )
    rc.set("warmup", b"1")
    print(f"[{args.label or 'run'}] connected, {len(rc.get_primaries())} primaries")

    threads = [threading.Thread(target=worker, args=(rc,), daemon=True) for _ in range(args.threads)]
    threads += [
        threading.Thread(
            target=chaos,
            args=(rc, args.container, ports, args.freeze, args.gap, args.max_frozen),
            daemon=True,
        )
        for _ in range(args.max_frozen + 1)
    ]
    for t in threads:
        t.start()

    started = time.monotonic()
    while time.monotonic() - started < args.duration and not STOP.is_set():
        time.sleep(0.5)
    STOP.set()
    for t in threads:
        t.join(timeout=10)

    elapsed = time.monotonic() - started
    print(f"\n[{args.label or 'run'}] {elapsed:.1f}s elapsed")
    for name, n in COUNTS.most_common():
        print(f"  {n:>8}  {name}")

    if CRASH:
        print(f"\n[{args.label or 'run'}] REPRODUCED — first traceback:\n")
        print(CRASH[0])
        return 1
    print(f"\n[{args.label or 'run'}] no RuntimeError in {elapsed:.1f}s")
    return 0


if __name__ == "__main__":
    sys.exit(main())
```

</details>

